PR #102, opened 2014-05-26
support seialized object with Key & Value method

Ray-Ni:
Add another way to serialize data


xerial:
Sorry. We no longer maintains msgpack-java v6. Let me close this PR

PR #168, opened 2015-01-03
RawStringValueImpl#equals and #hashcode use the same logics with StringV...

komamitsu:
...alueImpl

I thinks RawStringValueImpl#equals and #hashcode should use the same logics as StringValueImpl. I saw the performance of StringValueImpl#equals is two times better than RawStringValueImpl's one. So I applied StringValueImpl's logics to RawStringValueImpl.


xerial:
looks good. Thanks!

PR #169, opened 2015-01-10
MessageBufferOutput#reset shouldn't close an output passed as a paramete...

komamitsu:
...r

(`ChannelBufferOutput`|`OutputStreamBufferOutput`)#reset close WritableByteChannel or OutputStream passed as a parameter. As a result, when outputting to a file, it causes a write error because the file is already closed.

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/buffer/ChannelBufferOutput.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/buffer/ChannelBufferOutput.java
@@ -19,6 +19,7 @@ public class ChannelBufferOutput implements MessageBufferOutput {
     }
 
     public void reset(WritableByteChannel channel) throws IOException {
+        this.channel.close();
         this.channel = channel;
     }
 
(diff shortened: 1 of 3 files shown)


komamitsu:

msgpack-core/src/main/java/org/msgpack/core/buffer/ChannelBufferOutput.java
```diff
@@ -19,7 +19,6 @@ public ChannelBufferOutput(WritableByteChannel channel) {
     }
 
     public void reset(WritableByteChannel channel) throws IOException {
-        channel.close();
```
xerial: This code should be:
```
this.channel.close();
```

xerial:
My intention is to close the previously assigned output/channel (not the on in the input argument). So simply modifying the code to use `this.out.close()` is sufficient?

msgpack-core/src/main/java/org/msgpack/core/buffer/OutputStreamBufferOutput.java
```diff
@@ -19,7 +19,6 @@ public OutputStreamBufferOutput(OutputStream out) {
     }
 
     public void reset(OutputStream out) throws IOException {
-        out.close();
```
xerial: This should have be:
```
this.out.close();
```

komamitsu:
Sounds nice. I changed `reset()` again to close the previous output.

xerial:
thx

PR #172, opened 2015-01-12
Set "-Xdoclint:none" only when the Java is 1.8 or more

komamitsu:
When executing "publishLocal" on sbt, it failed with an error message `[error] javadoc: error - invalid flag: -Xdoclint:none`. I guess the reason why it failed is javadoc doesn't has the option in Java1.7 or before. If all the msgpack-java developers use Java1.8, we have nothing to do to fix this issue. But I don't think the limitation is so good. I fixed this issue but I'm not sure if this change is the best.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -59,7 +59,16 @@ object Build extends Build {
       scalacOptions ++= Seq("-encoding", "UTF-8", "-deprecation", "-unchecked", "-target:jvm-1.6", "-feature"),
       javaOptions in Test ++= Seq("-ea"),
       javacOptions in (Compile, compile) ++= Seq("-encoding", "UTF-8", "-Xlint:unchecked", "-Xlint:deprecation", "-source", "1.6", "-target", "1.6"),
-      javacOptions in doc := Seq("-source", "1.6", "-Xdoclint:none"),
+      javacOptions in doc := {
+        val opts = Seq("-source", "1.6")
+        val (major, minor) = System.getProperty("java.version").split('.') match {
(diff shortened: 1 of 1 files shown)


xuwei-k:
https://github.com/scala/scala/blob/v2.10.4/src/library/scala/util/Properties.scala#L159-L182

xerial:
@komamitsu 
Use scala.util.isJavaAtLeast as @xuwei-k suggested.

project/Build.scala
```diff
@@ -59,7 +59,16 @@ object Build extends Build {
       scalacOptions ++= Seq("-encoding", "UTF-8", "-deprecation", "-unchecked", "-target:jvm-1.6", "-feature"),
       javaOptions in Test ++= Seq("-ea"),
       javacOptions in (Compile, compile) ++= Seq("-encoding", "UTF-8", "-Xlint:unchecked", "-Xlint:deprecation", "-source", "1.6", "-target", "1.6"),
-      javacOptions in doc := Seq("-source", "1.6", "-Xdoclint:none"),
+      javacOptions in doc := {
+        val opts = Seq("-source", "1.6")
+        val (major, minor) = System.getProperty("java.version").split('.') match {
+          case Array(major, minor, _) => (major.toInt, minor.toInt)
+        }
+        if ((major == 1 && minor >= 8) || (major > 1))
```
xerial: `scala.util.Properties.isJavaAtLeast("1.8")`

komamitsu:
@xuwei-k @xerial Thanks. I replaced it with `scala.util.Properties.isJavaAtLeast()`

xerial:
thanks!

Issue #173, opened 2015-01-12
Reconsider calling close() of the previous assigned stream or channel in MessageBuffer(Input|Output)#reset


komamitsu:
In https://github.com/msgpack/msgpack-java/pull/169 and https://github.com/msgpack/msgpack-java/pull/170, we made some changes to call a previously assigned stream or channel instead of a new one passed as a parameter.

But I found https://github.com/spring-projects/spring-framework does the followings when it returns a HTTP request with jackson-dataformat-msgpack.
- Reuses an OutputStream (`org.apache.catalina.connector.CoyoteOutputStream`) without having called its close()
 - Of course, it includes a new `SocketChannel` each time
- Creates a new `org.msgpack.jackson.dataformat.MessagePackGenerator`
 - Which calls `org.msgpack.core.buffer.OutputStreamBufferOutput#reset` internally except for the first time 

As a result, the OutputStream the framework tries to reuse will be closed unexpectedly and MessagePack fails to write data to it.

Also, on second thought, I think MessageBuffer(Input|Output)#reset shouldn't close the previous output or channel for the following reasons.
- Closing the previous output or channel in the methods is unexpected behaviour for the user and we can't know what it causes (it depends on their implementation)
- Releasing those resources are basically the user's responsibility

Also, on second thought, it sounds MessageBuffer(Input|Output)#reset shouldn't close the previous output or channel for the following reasons.

@xerial What do you think? If the above ideas sound good, I can fix this issue easily.

xerial:
OK. Delegate the responsibility of closing the previous stream and channel to the caller of `reset` methods. I think changing the return value of `reset` to OutputStream (or OutputChannel) and never closing it inside `reset` method are sufficient.

komamitsu:
`MessagePacker` and `MessageUnpacker` have `reset()` method too that calls `MessageBuffer(Input|Output)#reset` internally. If we decide to make `MessageBuffer(Input|Output)#reset` return the previously assigned `(Input|Output)` Stream or Channel, `Message(Packer|Unpacker)#reset` should return it as well. But it needs more implementation. So I just removed the closing of the old `(Input|Output)` Stream or Channel in https://github.com/msgpack/msgpack-java/pull/175 for now.

Also, the integration of https://github.com/spring-projects/spring-framework and jackson-dataformat-msgpack work well with the PR.
https://github.com/komamitsu/SpringBootExample

komamitsu:
[...]

PR #175, opened 2015-01-15
Remove calling of the old resource in reset()

komamitsu:
For https://github.com/msgpack/msgpack-java/issues/173.

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
@@ -83,20 +83,17 @@ public class MessagePacker implements Closeable {
         this.position = 0;
     }
 
+    /**
+     * Reset output. This method doesn't close the old resource.
+     * @param out new output
+     */
     public void reset(MessageBufferOutput out) throws IOException {
         // Validate the argument
         MessageBufferOutput newOut = checkNotNull(out, "MessageBufferOutput is null");
 
-        try {
-            if(this.out != newOut) {
-                this.out.close();
-            }
-        }
-        finally {
-            // Reset the internal states here for the exception safety
-            this.out = newOut;
-            this.position = 0;
-        }
+        // Reset the internal states
+        this.out = newOut;
+        this.position = 0;
     }
 
 
--- a/msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
@@ -86,14 +86,17 @@ public class MessagePacker implements Closeable {
     /**
      * Reset output. This method doesn't close the old resource.
      * @param out new output
+     * @return the old resource
      */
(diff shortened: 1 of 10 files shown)


komamitsu:

msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
```diff
@@ -83,20 +83,17 @@ public MessagePacker(MessageBufferOutput out, MessagePack.Config config) {
         this.position = 0;
     }
 
+    /**
+     * Reset output. This method doesn't close the old resource.
+     * @param out new output
+     */
     public void reset(MessageBufferOutput out) throws IOException {
```
xerial: How about returning the old MessageBufferOutput instance here?

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
```diff
@@ -146,24 +146,21 @@ public MessageUnpacker(MessageBufferInput in, MessagePack.Config config) {
         this.config = checkNotNull(config, "Config");
     }
 
+    /**
+     * Reset input. This method doesn't close the old resource.
+     * @param in new input
+     */
     public void reset(MessageBufferInput in) throws IOException {
```
xerial: The same as above: returning the previous MessageBufferInput

msgpack-core/src/main/java/org/msgpack/core/buffer/ArrayBufferInput.java
```diff
@@ -24,6 +24,10 @@ public ArrayBufferInput(byte[] arr, int offset, int length) {
         this.buffer = MessageBuffer.wrap(checkNotNull(arr, "input array is null")).slice(offset, length);
     }
 
+    /**
+     * Reset buffer. This method doesn't close the old resource.
+     * @param buf new buffer
+     */
     public void reset(MessageBuffer buf) {
```
xerial: `public MesageBuffer reset(MessageBuffer buf)` ?

msgpack-core/src/main/java/org/msgpack/core/buffer/ByteBufferInput.java
```diff
@@ -16,6 +16,10 @@ public ByteBufferInput(ByteBuffer input) {
         this.input = checkNotNull(input, "input ByteBuffer is null");
     }
 
+    /**
+     * Reset buffer. This method doesn't close the old resource.
+     * @param input new buffer
+     */
     public void reset(ByteBuffer input) {
```
xerial: `public ByteBuffer reset(ByteBuffer input)`?

xerial:
@komamitsu 
Do you think it is useful to return the previous MessageBuffer etc. as the return value of `reset(...)` call? 

If not, simply using void is OK.

komamitsu:
@xerial Thanks. I fixed them again. Can you review it again?

xerial:
looks good. Thanks!

PR #179, opened 2015-01-17
Universal MessageBuffer implementation for supporting Java 6

### LINKED ISSUE #160: unsafe.copyMemory call error
I have a project running with JDK1.6 and msgpack v0.6.0,when I try to update to msgpack v0.7,it gaves me errors like:

The method copyMemory(long, long, long) in the type Unsafe is not applicable for the arguments (byte[], int, Object, long, int)	MessageBuffer.java

I google it and found it may cause by the copyMemory function with 5 params is  available on JDK 1.7 only(not sure) .If so,how can I manage this problem without update my JDK to 1.7? thx for helping!

xerial:
Some JDK in 1.6.x (and Android) does not have `unsafe.copyMemory(srcObj, srcOffset, dstObj, dstOffset, len)` method, which resolves the object memory address inside JVM, so we cannot replace this method without using JNI.

To fix this problem, I added an universal MessageBuffer implementation (MessageBufferU). This fixes #160.

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
@@ -30,9 +30,22 @@ public class MessageBuffer {
     static final boolean isByteBufferConstructorTakesBufferReference;
     static final int ARRAY_BYTE_BASE_OFFSET;
     static final int ARRAY_BYTE_INDEX_SCALE;
+    static final boolean isJavaAtLeast7;
 
     static {
         try {
+            // Check java version
+            String javaVersion = System.getProperty("java.specification.version", "");
+            int dotPos = javaVersion.indexOf('.');
+            if(dotPos == -1) {
+                isJavaAtLeast7 = false;
+            }
+            else {
+                int major = Integer.parseInt(javaVersion.substring(0, dotPos));
+                int minor = Integer.parseInt(javaVersion.substring(dotPos+1));
+                isJavaAtLeast7 = major > 1 || (major == 1 && minor >= 7);
+            }
+
             // Fetch theUnsafe object for Orackle JDK and OpenJDK
             Unsafe u;
             try {
@@ -92,7 +105,17 @@ public class MessageBuffer {
                     assert false;
             }
 
-            String bufferClsName = isLittleEndian ? "org.msgpack.core.buffer.MessageBuffer" : "org.msgpack.core.buffer.MessageBufferBE";
+            String bufferClsName;
+            if(isJavaAtLeast7) {
+                if(isLittleEndian)
+                    bufferClsName = "org.msgpack.core.buffer.MessageBuffer";
+                else
+                    bufferClsName = "org.msgpack.core.buffer.MessageBufferBE";
+            }
+            else {
+                bufferClsName = "org.msgpack.core.buffer.MessageBufferU";
+            }
+
             Class<?> bufferCls = Class.forName(bufferClsName);
             msgBufferClass = bufferCls;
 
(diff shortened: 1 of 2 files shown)


xerial:
@komamitsu 
Could you review this?  Thanks in advance.

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
```diff
@@ -472,12 +495,7 @@ public int offset() {
     }
 
     @Insecure
-    public Object getReference() { return reference; }
-
-
-    public void relocate(int offset, int length, int dst) {
```
xerial: Removed since `relocate` is never used.

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBufferU.java
```diff
@@ -0,0 +1,183 @@
...
+            resetBufferPosition();
+        }
+    }
+    @Override
+    public void putByte(int index, byte v) {
+        reference.put(index, v);
+    }
+    @Override
+    public void putBoolean(int index, boolean v) {
+        reference.put(index, v ? (byte) 1 : (byte) 0);
```
komamitsu: http://docs.oracle.com/javase/specs/jvms/se7/jvms7.pdf

> 2.3.4 The boolean Type
>    :
> In Oracle’s Java Virtual Machine implementation, boolean arrays in the Java programming language are encoded as Java Virtual Machine byte arrays, using 8 bits per boolean element.
> The Java Virtual Machine encodes boolean array components using 1 to represent true and 0 to represent false.

Looks good.

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBufferU.java
```diff
@@ -0,0 +1,183 @@
...
+    @Override
+    public void putByteBuffer(int index, ByteBuffer src, int len) {
+        assert (len <= src.remaining());
+
+        if(src.hasArray()) {
+            byte[] srcArray = src.array();
+            putBytes(index, srcArray, src.position(), len);
+
+        } else {
+            for(int i = 0; i < len; ++i) {
```
komamitsu: `putByte()` with loop sounds slow (I saw it was around 5x slower). How about using `java.nio.ByteBuffer#put(java.nio.ByteBuffer)` and `java.nio.Buffer#limit(int)` like this?

```
            int currLimitOfSrc = src.limit();
            src.limit(src.position() + len);
            reference.put(src);
            src.limit(currLimitOfSrc);
```

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
```diff
@@ -33,6 +33,19 @@
...
         try {
+            // Check java version
+            String javaVersion = System.getProperty("java.specification.version", "");
+            int dotPos = javaVersion.indexOf('.');
+            boolean isJavaAtLeast7 = false;
+            if(dotPos == -1) {
+                isJavaAtLeast7 = false;
+            }
+            else {
+                int major = Integer.parseInt(javaVersion.substring(0, dotPos));
```
komamitsu: We don't need to catch an exception from `parseInt()` or `substring()` just in case?

komamitsu:
Looks good.

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBufferU.java
xerial: Good point.

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
xerial: `java.specification.version` should be a dot-separted integers but some user might wrongly set a non-integer value to this. OK. I'll add a fallback.

PR #185, opened 2015-01-28
Fix wrong skipValue position count

miniway:
`readNextLength16` and `readNextLength32` already advances buffer position so `skipValue` should not advance position on Array and MapType.


xerial:
Good catch! Thanks.

PR #186, opened 2015-01-28
improve travis settings

xuwei-k:


xerial:
I didn't know travis can use caches. Great!

Unfortunately it looks like the container-based infrastructure (enabled with `sudo: false`) shows different performance characteristics and some performance tests didn't pass. 

Maybe we should increase the repetition number to make the performance test more stable.

xerial:
@xuwei-k 
Could you increase this number, say 30? 
https://github.com/msgpack/msgpack-java/blob/v07-develop/msgpack-core/src/test/scala/org/msgpack/core/MessageUnpackerTest.scala#L556-556

xerial:
OK. Tests worked well. 

Let's see the build time improvement next time.

Issue #188, opened 2015-01-29
MessageBuffer.wrap(ByteBuffer) is broken


superfell:
the MessageBuffer constructor that takes a ByteBuffer behaves wrongly if the supplied byte buffer only occupies a subset of the backing array, it ignores the bytebuffer subset, and just looks at the entire backing array. e.g. this test demonstrates the problem

  @Test
  public void testWrappedByteBuffer() {
    byte [] d = {10,11,12,13,14,15,16,17,18,19};
    ByteBuffer subset = ByteBuffer.wrap(d, 2, 2);
    MessageBuffer mb = MessageBuffer.wrap(subset);
    Assert.assertEquals(12, mb.getByte(0));  // fails returns 10
    Assert.assertEquals(2, mb.size()); // fails return 10
  }

The code seems to assume that ByteBuffer.array() returns an array of just the current Buffer contents, but in fact it returns the entire backing array.

xerial:
If you need to add an offset to the internal buffer, use MessageBuffer.slice.   MessageBuffer.wrap is used  for wrapping buffer types (e.g., byte[], ByteBuffer) with MessageBuffer.

xerial:
Fixed. Thanks for reporting the bug!

PR #192, opened 2015-02-02
#188: Fix for MessageBufferU

xerial:
Fix broken build #188


xerial:
@komamitsu Fixed MessageBufferU. And also add a jvm option `msgpack.universal-buffer=true` to force using MessageBufferU, which would be useful for testing it in Java7/8

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
```diff
@@ -119,7 +119,8 @@
             // We need to use reflection to find MessageBuffer implementation classes because
             // importing these classes creates TypeProfile and adds some overhead to method calls.
             String bufferClsName;
-            if(!isAndroid && isJavaAtLeast7) {
+            boolean useUniversalBuffer = Boolean.parseBoolean(System.getProperty("msgpack.universal-buffer", "false"));
```
komamitsu: :+1:

komamitsu:
Looks good

PR #193, opened 2015-02-04
Make MessagePacker internal state protected so it can be extended

slandelle:
Because MessagePacker internal state is private, there's no way to extend it.
This PR changes the visibility to protected. The other possibility would be to provide accessors for those.

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
@@ -55,15 +55,15 @@ public class MessagePacker implements Closeable {
     private final MessagePack.Config config;
 
     private MessageBufferOutput out;
-    private MessageBuffer buffer;
-    private MessageBuffer strLenBuffer;
+    protected MessageBuffer buffer;
+    protected MessageBuffer strLenBuffer;
 
-    private int position;
+    protected int position;
 
     /**
      * String encoder
      */
-    private CharsetEncoder encoder;
+    protected CharsetEncoder encoder;
 
 
     /**


xerial:
@slandelle Could you let me know for what purpose do you need to access these fields? `private field` is subject to change for future optimization, so if you write some code based on these field names, it would be difficult for you to maintain your project.

slandelle:
@xerial Some crazy hack.
I'm trying to insert the size of a packed payload just *before* the payload itself, without having to pack the payload separately and copy it back into the message itself.
I want to do that because I'd like to fetch the whole payload at once without unpacking when I'll be unpacking the message.

The hack I have in mind is to write an int as a placeholder for the size, pack the payload while keeping track of the placeholder position and the packed payload size, and override the placeholder.

As I said, crazy ugly hack. But still I think my use case makes sense.

slandelle:
@xerial I could maybe find another way to implement my use case, if I were to have a way to know how many bytes a MessagePacker has packed at any given point of time. But MessagePacker lacks a `totalReadBytes` just like MessageUnpacker has a `totalReadBytes`.

Would you consider a PR adding this feature?

xerial:
Adding totalWrittwnBytes sounds nice. Counting the total number of bytes adds some overhead, but MessagePacker is not so sensitive to the performance.

I'll welcome PR if you have time to create it.

slandelle:
@xerial rock'n roll. Do you have an ETA for 0.7.0-p6?

xerial:
At anytime. If you want it tomorrow, I can release it. Releasing a new version is 10 minutes work.

slandelle:
You're the man!

PR coming in a few minutes.

slandelle:
Closing this PR.

Issue #194, opened 2015-02-06
Supporting platform that has no sun.misc.Unsafe


xerial:
We have got an report that the current msgpack-core doesn't work in GAE. MessageBuffer implementation is designed to achieve the best performance for accessing memory through sun.misc.Unsafe. But Unsafe is not avilable in GAE.

 - Loading sun.misc.Unsafe might fail in GAE. 
  - Simply referencing it in the code (in import statement or field declaration) may have no problem; @xuwei-k reported the error log when using msgpack-core-0.7.0-p5 in https://gist.github.com/xuwei-k/31f9c9256439af0c3272  (error occurred after `Unsafe u;` statement)
  - Android has sun.misc.Unsafe so there was a workaround #180
 - We NEVER create any interface for MessageBuffer because it changes invokevirtual calls to invokeinterface and deteriorates the performance. And also having two or more classes that implement MessageBuffer methods generates TypeProfile. It makes each method call 2x slower.

xerial:
I'm working on this task on https://github.com/xerial/msgpack-java/tree/v07-no-unsafe branch. 

I have succeeded in using MessageBufferU, which does not rely on Unsafe class. Another problem is there is no java.nio.DirectByteBufffer in GAE. I'm seeking a workaround.

xerial:
Just deployed 0.7.0-p7, which supports GAE.

oza:
Good job!

PR #198, opened 2015-02-10
Google App Engine Support

xerial:
Fixes for #194: 
- Refactored MessageBuffer to isolate the initialization steps between standard and Java6/Android/GAE platforms. 
- Extracted DirectBufferAccess code to a new file. 

I confirmed with this fix, we can initialize MessageBuffer class in GAE platform.

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
@@ -5,12 +5,9 @@ import sun.misc.Unsafe;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
 import java.nio.BufferOverflowException;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
-import java.security.AccessControlException;
 
 import static org.msgpack.core.Preconditions.*;
 
@@ -26,17 +23,9 @@ public class MessageBuffer {
 
     static final boolean isUniversalBuffer;
     static final Unsafe unsafe;
-    // TODO We should use MethodHandle for efficiency, but it is not available in JDK6
+
     static final int ARRAY_BYTE_BASE_OFFSET;
     static final int ARRAY_BYTE_INDEX_SCALE;
-
-    enum DirectBufferConstructorType {
-        ARGS_LONG_INT_REF,
-        ARGS_LONG_INT,
-        ARGS_INT_INT,
-        ARGS_MB_INT_INT,
-        NONE
-    }
 
     static {
         try {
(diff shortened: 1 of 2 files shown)


xerial:
@komamitsu 
Could you review the changes? Thanks in advance.

xuwei-k:
I haven't been able to understand all changes, but I confirmed this code works on Google App Engine. 
nice work :+1: 

https://github.com/xuwei-k/msgpack-json/commit/00c289901a26396498f8ab

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
```diff
@@ -59,101 +44,62 @@
...
+                boolean isLittleEndian = ByteOrder.nativeOrder() == ByteOrder.LITTLE_ENDIAN;
+                if(isLittleEndian) {
+                    bufferClsName = "org.msgpack.core.buffer.MessageBuffer";
+                }
+                else {
+                    bufferClsName = "org.msgpack.core.buffer.MessageBufferBE";
+                }
 
+                // Fetch theUnsafe object for Oracle and OpenJDK
+                Unsafe u;
```
komamitsu: Minor: This variable isn't used

xerial: Removed. Thanks

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
```diff
@@ -59,101 +44,62 @@
...
             String bufferClsName;
-            boolean useUniversalBuffer = Boolean.parseBoolean(System.getProperty("msgpack.universal-buffer", "false"));
-            if(!useUniversalBuffer && !isAndroid && isJavaAtLeast7) {
-                if(isLittleEndian)
-                    bufferClsName = "org.msgpack.core.buffer.MessageBuffer";
-                else
-                    bufferClsName = "org.msgpack.core.buffer.MessageBufferBE";
-            }
-            else {
+            if(isUniversalBuffer) {
```
komamitsu: How about something like this to make it more robust?
```
try {
    if (!isUniversalBuffer) {
            :
        boolean isLittleEndian = ByteOrder.nativeOrder() == ByteOrder.LITTLE_ENDIAN;
         if(isLittleEndian) {
             bufferClsName = "org.msgpack.core.buffer.MessageBuffer";
         }
         else {
             bufferClsName = "org.msgpack.core.buffer.MessageBufferBE";
         }
    }
}
catch {
   // Let's use MessageBufferU
}
if (bufferClsName == null) {
    bufferClsName = "org.msgpack.core.buffer.MessageBufferU";
        :
}
```

xerial: Good idea. With this fix, msgpack-java works anyway. To do this, I need to put the initialization code of static fields (bufferClsName, unsafe, etc.) after the last catch call.

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
```diff
@@ -351,6 +252,9 @@ else if(directByteBufferClass.isInstance(buffer.base)) {
      */
     MessageBuffer(ByteBuffer bb) {
         if(bb.isDirect()) {
+            if(isUniversalBuffer) {
+                throw new IllegalStateException("Cannot create MessageBuffer from DirectBuffer");
```
komamitsu: This code doesn't seem to be called since the constructor of MessageBufferU will be called if `isUniversalBuffer` is true.

xerial: Right. This code should not be called and that's why I throw IllegalStateException.

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
xerial: Fixed in 5fd6964

komamitsu:
Looks good

komamitsu:
It worked on Android 2.3.3, 4.0.4 and 5.0 :)

PR #201, opened 2015-02-25
jackson: Add support for Extended type

ramonmaruko:
Please review my code.

### DIFF
--- a/msgpack-jackson/src/test/java/org/msgpack/jackson/dataformat/MessagePackParserTest.java
+++ b/msgpack-jackson/src/test/java/org/msgpack/jackson/dataformat/MessagePackParserTest.java
@@ -7,6 +7,7 @@ import org.junit.Test;
 import org.msgpack.core.MessagePack;
 import org.msgpack.core.MessagePacker;
 import org.msgpack.core.buffer.OutputStreamBufferOutput;
+import org.msgpack.value.ExtendedValue;
 
 import java.io.*;
 import java.math.BigInteger;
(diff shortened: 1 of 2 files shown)


msgpack-jackson/src/test/java/org/msgpack/jackson/dataformat/MessagePackParserTest.java
```diff
@@ -128,14 +135,19 @@ else if (k.equals("bool")) {
                 bitmap |= 1 << 9;
                 assertEquals(false, v);
             }
+            else if (k.equals("ext")) {
+                // #9
+                bitmap |= 1 << 10;
+                assertArrayEquals(extPayload, ((ExtendedValue) v).toByteArray());
```
komamitsu: How can we get the type of this extended type(2) ?

komamitsu: Oh, sorry we can use org.msgpack.value.ExtendedValue#getExtType. never mind.

komamitsu:
Looks good. Thanks for the pull-request.

komamitsu:
@ramonmaruko On second thought, it think `org.msgpack.jackson.dataformat.MessagePackParser#getEmbeddedObject` should return `org.msgpack.jackson.dataformat.MessagePackExtendedType` (doesn't exist yet) or something which has a extended type and payload instead of `org.msgpack.value.ExtendedValue directly`. I found other MessagePackParser's APIs don't return `org.msgpack.value.ValueRef`. Also, if the users get `org.msgpack.value.ExtendedValue` directly, they need to change their code when the implementation of org.msgpack.value.ExtendedValue is changed. What do you think?

PR #203, opened 2015-03-07
Make MessagePackParser#getEmbeddedObject return MessagePackExtendedType

komamitsu:
In https://github.com/msgpack/msgpack-java/pull/201, we made `org.msgpack.jackson.dataformat.MessagePackParser#getEmbeddedObject` return `org.msgpack.value.ExtendedValue` directly. But `org.msgpack.value.ExtendedValue` shouldn't be exposed from `org.msgpack.jackson.dataformat`, I think. Also, wrapping it with `MessagePackExtendedType` would be useful when the implementation of `org.msgpack.value.ExtendedValue` changes.


ramonmaruko:
:thumbsup:

PR #205, opened 2015-03-13
Added support for BigDecimal, fixes #204

### LINKED ISSUE #204: Support for BigDecimal in Jackson Data Format
Currently there is no support to read/write java BigDecimals. This can be done in two ways:

1. Encode BigDecimal to String (accurate (math-wise), not portable to other langs, inefficient)

2. Encode to Integer or Float depending on the value and warn in case of overflow/underflow (not accurate (math-wise), portable to other langs, efficient)

I prefer option 2, since that's already what MessagePack does with BigInteger.

kutchar:
Added support for BigDecimal, fixes #204 using option 2


komamitsu:
Great. Thank you!

kutchar:
You're welcome. Any idea when this would be released?

komamitsu:
This will be released after fixing https://github.com/msgpack/msgpack-java/issues/209, I think.

PR #207, opened 2015-03-13
Fixing the contract of nextToken to be compliant with com.fasterxml.jackson.core.JsonParser.nextToken(), fixes #206

### LINKED ISSUE #206: MessagePackParser doesn't follow the JsonParser contract for nextToken
Based on the javadoc of `com.fasterxml.jackson.core.JsonParser.nextToken()`

```
/**
 * Main iteration method, which will advance stream enough
 * to determine type of the next token, if any. If none
 * remaining (stream has no content other than possible
 * white space before ending), null will be returned.
 *
 * @return Next token from the stream, if any found, or null
 *   to indicate end-of-input
 */
public abstract JsonToken nextToken() throws IOException, JsonParseException;
```

`nextToken` should return `null` if stream has no more content, but currently it throws an `EOFException`.

kutchar:
Fixing the contract of nextToken to be compliant with com.fasterxml.jackson.core.JsonParser.nextToken() and not throw an EOFException if at the end of the stream. fixes #206

### DIFF
--- a/msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackParser.java
+++ b/msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackParser.java
@@ -118,6 +118,10 @@ public class MessagePackParser extends ParserMinimalBase {
             }
         }
 
+        if (!messageUnpacker.hasNext()) {
+            return null;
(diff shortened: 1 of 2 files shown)


komamitsu:
Good catch! Thanks

PR #212, opened 2015-03-17
Make MessageBufferU constructor public, close #210

### LINKED ISSUE #210: Make MessageBufferU constructor public
Some APIs, such as Akka's ByteString, only provide read-only ByteBuffers. It would be great to make MessageBufferU constructor public instead of package private so I can easily write a custom MessageInput in my own package.

Regards

slandelle:
So it can be used to unpack read-only ByteBuffers


slandelle:
The test fails on Travis because it runs with JDK6. The test assumes JDK7 to demonstrate that a read-only ByteBuffer can't normally be read, as it elects regular MessageBuffer implementation. But with JDK6, it elects MessageBufferU.

Fixing.

slandelle:
Test fixed. I sadly had to move the test to the `org.msgpack.core.buffer` package, as `MessageBuffer`'s constructor is package private too, just like `MessageBufferU` was until this PR.

project/Build.scala
```diff
@@ -165,7 +165,8 @@ object Build extends Build {
       "org.xerial" % "xerial-core" % "3.3.5" % "test",
       "org.msgpack" % "msgpack" % "0.6.11" % "test",
       junitInterface,
-      "commons-codec" % "commons-codec" % "1.10" % "test"
+      "commons-codec" % "commons-codec" % "1.10" % "test",
+      "com.typesafe.akka" %% "akka-actor"% "2.3.9"
```
xerial: use `test` scope for akka.

slandelle: facepalm

slandelle: Fixed. Do you want me to squash the commits?

xerial:
No need to squash the commits. They are small enough

xerial:
Looks good.

slandelle:
Thanks!

Issue #209, opened 2015-03-13
packString with a string which is larger than 8192 bytes fails to unpack data


oza:
Following test fails:
```
  def testLargeData : Array[Byte] = {
    val out = new ByteArrayOutputStream()
    val packer = msgpack.newPacker(out)
    val stringLength = 8193

    packer
      .packArrayHeader(2)
      .packString("l" * stringLength)
      .packInt(1)

    packer.close()

    val arr = out.toByteArray
    debug(s"packed: ${toHex(arr)}, size:${arr.length}")

    arr
  }
```

```
    "parse message large packed data" taggedAs("unpack") in {
      val arr = testLargeData

      val unpacker = msgpack.newUnpacker(arr)

      var count = 0
      while(unpacker.hasNext) {
        count += 1
        readValue(unpacker)
      }
      count shouldBe 3
      unpacker.getTotalReadBytes shouldBe arr.length
    }
```

If stringLength is 8192, the test passes.

oza:
@xerial @komamitsu thanks for your fix!

PR #215, opened 2015-04-04
Fix MessageUnpacker#unpackString in case that string length is more than...

komamitsu:
I modified MessageUnpacker#unpackString to fix https://github.com/msgpack/msgpack-java/issues/209.

@xerial Can you review this change?

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
@@ -980,7 +980,6 @@ public class MessageUnpacker implements Closeable {
 
                         if(cr.isOverflow()) {
                             // The output CharBuffer has insufficient space
-                            readLen = bb.limit() - bb.remaining();
                             decoder.reset();
                         }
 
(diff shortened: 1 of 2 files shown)


xerial:
I found another bug; This part should use `ensure(strLen-cursor)`:
```
    while(cursor < strLen) {
                     if (!ensure(strLen))
                         throw new EOFException();
```

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
```diff
@@ -1005,9 +1004,10 @@ public String unpackString() throws IOException {
                         sb.append(decodeBuffer);
 
                         decodeBuffer.clear();
-                        cursor += readLen;
-                        consume(readLen);
                     }
+
+                    cursor += readLen;
```
xerial: I think there is no guarantee that the read length from the input always matches `Math.min(buffer.size() - position, strLen-cursor)`  when OVERFLOW (output buffer has insufficient space) happens.  In this case, the input data is partially read. That is why I compute `readLen` as `bb.limit() - bb.remaining()`.

xerial: Ah. I was wrong. `readLen` can be retrieved by `bb.position()` since we created a new ByteBuffer at 
```
ByteBuffer bb = buffer.toByteBuffer(position, readLen);
```
whose initial position is 0.

xerial:
Now I got it. I don't need to set `readLen` in `while(bb.hasRamaining())` loop. Thanks

PR #216, opened 2015-04-08
Fix buffer size to ensure

xerial:


msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
```diff
@@ -965,7 +965,7 @@ public String unpackString() throws IOException {
                 decodeBuffer.clear();
                 StringBuilder sb = new StringBuilder();
                 while(cursor < strLen) {
-                    if (!ensure(strLen))
+                    if (!ensure(strLen-cursor))
                         throw new EOFException();
 
                     int readLen = Math.min(buffer.size() - position, strLen-cursor);
```
komamitsu: `ensure(int byteSizeToRead)` looks to guarantee to prepare a buffer requested as `byteSizeToRead`. So `readLen` should be simply `strLen - cursor`? If so, maybe we can remove the outer loop.

xerial: looks like so. i'll remove it

xerial:
@komamitsu 
Fixed to utilize existing buffer rather than ensuring the buffer size for the whole string size.

komamitsu:
LGTM

PR #220, opened 2015-04-27
Automate sbt release steps

xerial:
This fix adds `publishSigned` and `sonatypeRelease` in `sbt release` steps.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -22,8 +22,11 @@ import de.johoop.findbugs4sbt.FindBugs._
 import de.johoop.jacoco4sbt._
 import JacocoPlugin._
 import sbtrelease.ReleasePlugin._
+import sbtrelease.ReleaseStateTransformations._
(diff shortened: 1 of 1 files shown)


project/Build.scala
```diff
@@ -55,6 +57,29 @@ object Build extends Build {
...
+        checkSnapshotDependencies,
+        inquireVersions,
+        runClean,
+        runTest,
+        setReleaseVersion,
+        commitReleaseVersion,
+        tagRelease,
+        ReleaseStep(
+          action = { state =>
+            val extracted = Project extract state
```
komamitsu: [Minor] Why don't you call `state.extract`?

xerial: state.extract is necessary to execute `runAggregated` task.

xerial: And we cannot call `state.extract`. We need to call `Project.extract(configuration key)`

komamitsu:
Looks good.

PR #224, opened 2015-05-01
Fixes #223: MapValueImpl does not implement getMapCursor

### LINKED ISSUE #223: MapValueImpl does not implement getMapCursor
MapValueImpl does not override the getMapCursor function and throws saying it is not a map

    ValueRef value = MapValueImpl.empty();
    System.out.println(value.getValueType());
    value.getMapCursor(); // throws...

Output

    MAP
    Exception in thread "main" org.msgpack.core.MessageTypeException: This value is not a map type

fdinoff:


xerial:
Thanks!

PR #225, opened 2015-05-03
Upgrade to sbt-sonatype-0.5.0

xerial:
Upgrade to sbt-sonatype-0.5.0 
- This version improves error handling on 500 error when accessing sonatype REST API. 
- Extract sbt-sonatype related setting into sonatype.sbt
- Enables sonatype related commands (e.g., sonatypeRelease) only at the root project
- publishTo setting is not longer necessary (sbt-sonatype sets this)


komamitsu:
Looks good

PR #237, opened 2015-05-25
Fixes #222: Check available() before calling blocking read

### LINKED ISSUE #222: InputStreamBufferInput.next() hangs with a blocking InputStream.
Version: 0.7.0-p9

I'm trying to unpack a message that is stored in a socket. An example is shown below so that you can easily replicate this.

```
import org.msgpack.core.MessagePack;
import org.msgpack.core.MessagePacker;
import org.msgpack.core.MessageUnpacker;

import java.io.IOException;
import java.net.ServerSocket;
import java.net.Socket;

public class BlockingTest {
    public static void main(String[] args) {

        Thread server = new Thread(new Runnable() {
            @Override
            public void run() {
                try (ServerSocket listener = new ServerSocket(9090)){
                    try (Socket socket = listener.accept()) {
                        MessagePack msgPack = new MessagePack();
                        MessagePacker packer = msgPack.newPacker(socket.getOutputStream());
                        packer.packArrayHeader(2).packLong(1).packString("Hello World");
                        packer.flush();
                        while (true) {
                            // Server leaves socket open for further messages
                        }
                    }
                } catch (IOException e) {}
            }
        });
        server.start();

        Thread client = new Thread(new Runnable() {
            @Override
            public void run() {
                try {
                    try (Socket socket = new Socket("localhost", 9090)) {
                        MessageUnpacker unpacker =
                                new MessagePack().newUnpacker(socket.getInputStream());
                        System.out.println(
                                "Socket Open with " + socket.getInputStream().available());
                        unpacker.hasNext(); // Hangs Here
                        System.out.println("this will never print");
                    }
                } catch (IOException e) {
                    System.out.println("failed");
                }
            }
        });
        client.start();
    }
}
```

The output is 

    Socket Open with 14

Currently `hasNext()` hangs since `InputStreamBufferInput.next()` tries to read from the InputStream looking for an EOF or fill its internal buffer which appears to default to 8192. However it will not do either of these since the socket never closes and the socket currently only has 14 bytes in it.

What is the proper way to decode messages from a blocking InputStream?

xerial:
For #222

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/buffer/InputStreamBufferInput.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/buffer/InputStreamBufferInput.java
@@ -54,24 +54,13 @@ public class InputStreamBufferInput implements MessageBufferInput {
         if(reachedEOF)
             return null;
 
-        byte[] buffer = null;
-        int cursor = 0;
-        while(!reachedEOF && cursor < bufferSize) {
-            if(buffer == null) {
-                buffer = new byte[bufferSize];
-            }
-
-            int readLen = -1;
-            // available() == 0 means, it reached the end of the stream
-            if(in.available() == 0 ||
-                (readLen = in.read(buffer, cursor, bufferSize - cursor)) == -1) {
-                reachedEOF = true;
-                break;
-            }
-            cursor += readLen;
+        byte[] buffer = new byte[bufferSize];
+        int readLen = in.read(buffer);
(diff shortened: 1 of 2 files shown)


xerial:
@komamitsu 
Could you review this?  Without this fix, the newly added test code hangs.

xerial:
@komamitsu 
Now fixed as @fdinoff suggested. Could you check this?

komamitsu:
Looks good

PR #240, opened 2015-05-27
Merge v07-develop to master

xerial:
This PR makes v07-develop the default branch (master): 
- Created v06 branch based on the current master, to maintain 0.6.x series 
- Forced merge (overwrite) files in master (0.6.x) with 0.7.x (v07-develop) files


komamitsu:
@xerial This PR will overwrite all the history of v07 code set.
```
[komamitsu@potato msgpack-java (master)]$ git log msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackFactory.
java
commit b25778b63eb5d0a874a9997f7accc7ea8969ddd5
Author: Taro L. Saito <leo@xerial.org>
Date:   Wed May 27 13:25:24 2015 +0900

    Force merge v07-develop to master
[komamitsu@potato msgpack-java (master)]$
```
So I don't think it's the best way. How about renaming remote branches? http://stackoverflow.com/questions/1526794/rename-master-branch-for-both-local-and-remote-git-repositories

xerial:
Good point. I didn't notice that this does not include all of the history of v07.

xerial:
I'll try rebasing so that we can keep both history of v06 and v07.

xerial:
Rebasing doesn't work as expected. So renaming is better. 
When renaming branches, several pull requests for 0.6.x (current master) need to be re-created to point  v06 branch.

xerial:
We should work on this issue after cleaning up old pull requests to 0.6.x

PR #245, opened 2015-06-07
Throw IOException in MessagePackParser close instead of printing

fdinoff:
Let the caller handle the exception instead of printing


xerial:
OK.

PR #246, opened 2015-06-10
V07 value impl

frsyuki:
This pull request implements Value API.


msgpack-core/src/main/java/org/msgpack/value/impl/ImmutableArrayValueImpl.java
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    @Override
+    public Value get(int index) {
+        return array[index];
+    }
+
+    @Override
+    public Value getOrNilValue(int index) {
+        if (array.length < index && index >= 0) {
```
fdinoff: This condition seems to be backwards. I think it should be `index < array.length`

frsyuki: ouch. fixed.

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
```diff
@@ -531,57 +525,143 @@ public void skipValue() throws IOException {
...
+                return ValueFactory.newBooleanValue(unpackBoolean());
+            case INTEGER:
+                switch (mf) {
+                    case UINT64:
+                        return ValueFactory.newIntegerValue(unpackBigInteger());
+                    default:
+                        return ValueFactory.newIntegerValue(unpackLong());
+                }
+            case FLOAT:
+                return ValueFactory.newFloatValue(unpackDouble());
```
xerial: I remembered why FloatHolder was necessary. When reading a float value as double, it will add some fractions. For example: 
```
scala> 0.1341f.toDouble
res5: Double = 0.13410000503063202
```
If user wants to print this float value as String, `0.13410000503063202` will be shown. This is unexpected behavior for the user.

Another example: `0.1f` becomes '0.10000000149011612` if we use double as `0.1f`s internal representation.

I'll fix this in another pull request.

frsyuki: I don't think it's a big deal. I mean we can say that it's a limitation because of internal representation of a Value. Anyways it is impossible to preserve all information of floating point numbers using decimal string format. We should accept that toString of Float values is best-effort.

If we create ImmutableFloatValueImpl,

1) Variable needs another float field which causes overhead and a lot more complexity.
2) Implementation of Value#equals becomes unclear.

I think benefit to have ImmutableFloatValueImpl is smaller than those negative effects.

If you think we should fix it, I suggest following solutions:

c) Always use ImmutableFloatValue always uses Float.toString. Give up double precision when users use json (decimal) string format.
a) Add `private final boolean singlePrecision" field to ImmutableFloatValue (and Variable) class. Only toString method considers it. If it is true, ImmutableFloatValueImpl#toString uses Float.toString.
b) Delete `ValueFactory#newFloatValue(float)` so that users can't create a Value with 0.134f at least in Java.

xerial: Discussed this topic with @frsyuki. MessagePack type system will not preserve the error range of float values since some msgpack implementation cannot have float values. So application developers needs to be careful so as not to write a code that depends on float precision. Even if the data is written in FLOAT32, Value interface will represent it in double precision.

fdinoff: Going this route means that if a single precision number is in the data stream and read into a value. `Value.writeTo` will write a double to the MessagePacker. Should `writeTo` reproduce the original msgpacked data or just something close to it?

xerial: This discussion has started in MessagePack specification side: 
https://github.com/msgpack/msgpack/pull/200

If we can preserve original message-packed data in an efficient way, lossless `writeTo` implementation would be useful, but the message type system does not require it.

xerial:
@muga 
I also feel `getXXX` and `toXXX` is confusing. And also `toString` produces JSON string that produces double quoted strings. There are lot of confusing points. 

How about renaming these methods as follows?: 
- (round)toXXX         (with truncation or rounding)
- (cast)asXXX throws Exception  (checked cast)
- toJSON   (produces json string. current implementation of toString)

Then remove getXXX methods.

frsyuki:
How about this simply?

* rename NumberValue#toXxx() methods to NumberValue#roundToXxx()

I think it's a possible idea to separate toJson from toString. However, because some values are not compatible with JSON (non-string keys of a map, ExtensionValue, floating point values, etc.), we would need another class like ValueToJsonPacker that converts Value to JSON.

xerial:
@frsyuki 
If we recommend to use Value.getXXX to extract values, simply renaming toXXX -> roundToXXX is better.

My concern on StringValue.toString is it always returns double quoted value in generating string messages with StringValue.

xerial:
Using double quoted string looks good for distinguishing StringValue "1" and FloatValue 1, for debugging. 

OK. I'll go with roundToXXX, getXXX and toString. toJSON should be implemented in another class as @frsyuki said.

xerial:
A problem is getting float value has no good name like roundToXXX. 

How about using `asXXX` for truncation and rounding (e.g., asInt, asFloat, etc., rather than roundToInt,  roundToFloat?). 

If we use asXXX, we can simply say representing this value as XXX type.

xerial:
Merged. Will add test codes and further internal improvement in another PR.

frsyuki:
Why don't you use roundToXxx? I think that castAsXxx assumes that users understand the actual implementation the methods. If we assume that users don't have to know the actual implementation, I think castToXxx is confusing because I don't know whether it throws exceptions or not. For example, `Double v = 1.1; Integer iv = (Integer) v;` throws an exception in Java. So, "cast" doesn't always implies rounding.

In my opinion, roundToFloat / roundToDouble is correct name because it actually can rounding.

frsyuki:
@xerial How about this radical idea?:

* Concept:
  * toXxx and asXxx are the type casting methods.
  * toXxx methods may lose some information but never throws exceptions.
  * asXxx methods don't lose information but could throw exceptions. Perhaps it's good idea to make them checked-exception so that difference of toXxx and asXxx becomes very clear.
* Implementation:
  * a) change IntegerValue#getXxx() to `asXxx()`
  * b) change NumberValue#castAsXxx() to `toXxx()` including `toFloat()` and `toDouble()`
  * c) change RawValue#getString() to `asString()`
  * d) remove StringValue#stringValue()
  * e) change StringValue#toString() so that it doesn't add double quotes
  * f) add `String Value#toJson()`. This implementation is same with current `#toString()`. So, it's easy to implement.
  * g) add `void Value#toJson(JsonPacker)` in the future so that we can customize behavior and improve performance if necessary. For example, "how to convert BinaryValue to JSON?", "how to convert ExntesionValue to JSON?", "how to convert FloatValue to JSON?", "how to handle non-string keys in a MapValue?"

With this idea, you can achieve short names and consistent concept at the same time.

xerial:
@frsyuki 
If we do f) (toString -> toJson), I agree with your idea:
  - toXXX (may lose information with casting, but throws no exception)
  - asXXX (may throw exception)

xerial:
@frsyuki 
But `NumberValue.asXXX` is already used for `Valuue.asXXXValue`. That was why I choose `castAsXXX` to distinguish methods that produce native values and msgpack Values.

frsyuki:
@xerial I think that Value.asAbcValue methods are also a kind of asXxx methods because it doesn't lose information, may throw exception, and Xxx is the returning class name.
(I edited above comment. toJson is not f but h. please check on github)

xerial:
If having `asInt` and `asIntegerValue` is OK, there would be no problem.

PR #249, opened 2015-06-13
V07 value impl revision 2

xerial:
This PR includes several API changes:
 - ExtendedType -> ExtensionType
 - ValueFactory.newXXXValue -> newXXX
    - Even if we remove this suffix, creating Value objects is evident since we are using ValueFactory. ValueFactory is a utility class for the ease of creating value objects. So shortening these method names is better for the developers. 
 - Renamed IntegerValue.integerValue, booleanValue to IntegerValue.toInt, toBoolean, etc. to avoid confusion; It is a little bit hard to distinguish whether we are creating Value or primitive type data from the context (e.g., `unpackValue(var); var.asIntegerValue().integerValue`) 
 - Removed `public` from interface definitions, since `public` is the default scope 

I'll work on adding test cases, ~~proper treatment of float/double values~~, etc. in the next pull request.


xerial:
@komamitsu 
Could you take a look at jackson test case failures? I fixed test cases in msgpack-core.

msgpack-core/src/main/java/org/msgpack/value/ValueFactory.java
```diff
@@ -124,17 +119,67 @@ ImmutableMapValue newMapValue(Map<K, V> map) {
...
     }
 
-    public static ImmutableExtendedValue newExtendedValue(byte type, byte[] data) {
-        return new ImmutableExtendedValueImpl(type, data);
+    public static ImmutableMapValue emptyMap() {
+        return ImmutableMapValueImpl.empty();
+    }
+
+
+    public static class MapEntry {
```
frsyuki: We don't need MapEntry class because we can use `Map.Entry<Value, Value>`.

msgpack-core/src/main/java/org/msgpack/value/ValueFactory.java
```diff
@@ -124,17 +119,67 @@ ImmutableMapValue newMapValue(Map<K, V> map) {
...
+        return new MapBuilder();
+    }
+
+    public static MapEntry newMapEntry(Value key, Value value) {
+        return new MapEntry(key, value);
+
+    }
+
+    public static class MapBuilder {
+        private Map<Value, Value> map = new HashMap<Value, Value>();
```
frsyuki: this can be `final`

msgpack-core/src/main/java/org/msgpack/value/ValueFactory.java
```diff
@@ -124,17 +119,67 @@ ImmutableMapValue newMapValue(Map<K, V> map) {
...
+        public MapBuilder() {}
+
```
frsyuki: It's good idea to add `putAll(Iterable<? extends Map.Entry<? extends Value,? extends Value>> entries)` and `putAll(Map<? extends Value,? extends Value> map)` because Guava's ImmutableMap.Builder has those methods.

msgpack-core/src/main/java/org/msgpack/core/ExtensionTypeHeader.java
```diff
@@ -3,16 +3,21 @@
...
  */
-public class ExtendedTypeHeader {
+public class ExtensionTypeHeader {
     private final byte type;
     private final int length;
 
-    ExtendedTypeHeader(byte type, int length) {
-        checkArgument(length >= 0, String.format("length must be >= 0: %,d", length));
-        this.length = length;
+    public ExtensionTypeHeader(byte type, int length) {
```
frsyuki: Why don't you add java doc here to suggest that users can use checkedCastToByte?

msgpack-core/src/main/java/org/msgpack/value/ValueFactory.java
xerial: Right. Fixed to use Map.Entry

msgpack-core/src/main/java/org/msgpack/value/ValueFactory.java
xerial: fixed

msgpack-core/src/main/java/org/msgpack/value/ValueFactory.java
xerial: Implement MapBuilder, following the interface of Guava.

msgpack-core/src/main/java/org/msgpack/core/ExtensionTypeHeader.java
xerial: Added a comment

komamitsu:
@xerial Fixed some errors in jackson-dataformat-msgpack https://github.com/xerial/msgpack-java/pull/1

PR #251, opened 2015-06-14
Fix MessageBufferU#putByteBuffer

komamitsu:


komamitsu:
@xerial Can you review this pull req?

xerial:
Thanks for the fix!

PR #254, opened 2015-06-23
Add universal buffer test instead of using jdk6

xerial:
- Fixes .travis.yml config not to use jdk6. This is necessary to run the latest version of checkstyle 
- Add test that uses MessageBufferU for emulating jdk6 and Android

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,6 @@
 language: scala
-scala:
-   - 2.11.1
 
 jdk:
-  - openjdk6
   - openjdk7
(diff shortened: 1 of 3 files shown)


komamitsu:
Looks good. Each JDK tests MessageBufferU though.

xerial:
Maybe we can drop one of oraclejdk7 or openjdk7 test

xerial:
TravisCI already uses 15 or more minutes to finish the tests. We may need some work to make it faster.

PR #255, opened 2015-06-23
Fix code style with sbt-jcheckstyle

xerial:
- Use facebook's coding style, which is closer to ours
- Enforce running jcheckStyle command before compile steps so that TravisCI can check coding style errors


komamitsu:
Looks good

PR #258, opened 2015-06-23
Fixes #38: Add IntegerValue.mostSuccinctMessageFormat

### LINKED ISSUE #38: IntegerValue should capture the size of the integer
The MessagePack serialized format knows the size of an Int, whether it's 8, 16, 32, or 64 bits.  However the current IntegerValue class loses this information, and it forces the user to pick the right method, getShort, getInt, etc.   It would be nice for IntegerValue to capture the original size of the integer when deserializing, and the same thing when it serializes.

xerial:
Add a utility method to IntegerValue to detect the most succinct integer type. This is useful when user wants to know which get method (e.g., getByte, getInt, etc. ) should be used to extract values.

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
@@ -423,7 +423,7 @@ public class MessagePacker
             writeByteAndLong(UINT64, bi.longValue());
         }
         else {
-            throw new IllegalArgumentException("Messagepack cannot serialize BigInteger larger than 2^64-1");
+            throw new IllegalArgumentException("MessagePack cannot serialize BigInteger larger than 2^64-1");
         }
         return this;
     }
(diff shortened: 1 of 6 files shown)


xerial:
@oza Could you review this change? I think this code is similar to your previous contribution to v06 (#38)

oza:
@xerial LGTM overall. I have a question about a test: Does forAll method generate values based on the types at random?
http://doc.scalatest.org/2.2.4/#org.scalatest.prop.GeneratorDrivenPropertyChecks

If the answer is positive, we should add a test with values on the borders of types explicitly. Thoughts?

xerial:
A nice thing about `forAll` method is it produces border values (e.g., Int.MaxValue, MinValue, 0, etc.) as well.

oza:
OIC. +1, merging this shortly.

oza:
Thanks @xerial for your contribution!

PR #264, opened 2015-06-30
Rename Extended to Extension

fdinoff:
@komamitsu Can you look at this? #250, and #262 used extended instead of extension. This renames the few variables and methods that used it.


xerial:
Looks good.

Issue #259, opened 2015-06-27
Rename data access methods


xerial:
Follow up of #246.
- toXXX (may lose information with casting, but throws no exception)
- asXXX (may throw exception)
- toJson  (produces json like string)

xerial:
Done by #265

PR #265, opened 2015-06-30
#259: Rename data access methods

xerial:


msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackParser.java
```diff
@@ -202,7 +202,7 @@ public JsonToken nextToken()
             case STRING:
                 value = messageUnpacker.unpackValue(var);
                 if (parsingContext.inObject() && _currToken != JsonToken.FIELD_NAME) {
-                    parsingContext.setCurrentName(value.asRawValue().stringValue());
+                    parsingContext.setCurrentName(value.asRawValue().toJson());
```
fdinoff: I'm surprised this works. I would have thought that the `toJson` would have quoted the string (as `ImmutableRawValueImpl` does). You actually get a `Variable$StringValueAccessor` which does't quote the string.

This depends on the output format of `toJson` however you wrote in the [java doc][1] 

> Do not write code that depends on the resulting json format.

but this depends directly on the output. Should this be using `toJson` or a method that would be more predictable?

   [1]: https://github.com/msgpack/msgpack-java/pull/265/files#diff-1f56e7353af89c0830ab0ade1552c7e2R253

msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackParser.java
```diff
@@ -212,7 +212,7 @@ public JsonToken nextToken()
             case BINARY:
                 value = messageUnpacker.unpackValue(var);
                 if (parsingContext.inObject() && _currToken != JsonToken.FIELD_NAME) {
-                    parsingContext.setCurrentName(value.asRawValue().stringValue());
+                    parsingContext.setCurrentName(value.asRawValue().toJson());
```
fdinoff: Same as above

msgpack-core/src/main/java/org/msgpack/value/RawValue.java
```diff
@@ -50,12 +50,6 @@
...
      */
-    String getString();
+    String asString();
 
-    /**
-     * Returns the value as {@code String}.
-     * <p/>
-     * This method replaces an invalid UTF-8 byte sequence with <code>U+FFFD replacement character</code>.
-     */
-    String stringValue();
```
frsyuki: I think it's better to keep `toString()` to override JavaDoc.

msgpack-core/src/main/java/org/msgpack/value/Variable.java
```diff
@@ -187,9 +187,15 @@ public int hashCode()
...
+        public String toJson()
+        {
+            return Variable.this.toJson();
+        }
+
+        @Override
         public String toString()
         {
-            return Variable.this.toString();
+            return toJson();
```
frsyuki: Do you mean `Variable.this.toString(); `?

msgpack-core/src/main/java/org/msgpack/value/Variable.java
```diff
@@ -678,7 +684,7 @@ public String getString()
         }
 
         @Override
-        public String stringValue()
+        public String toJson()
```
frsyuki: Isn't this `toString()`?

frsyuki:
@xerial I created this pull-request: https://github.com/xerial/msgpack-java/pull/2

xerial:
@frsyuki 
Thanks for the fix to produce json string in all of toJson() methods.

xerial:
@komamitsu 
Could you help me to fix the failed tests? 

Thanks.

komamitsu:
@xerial Created https://github.com/xerial/msgpack-java/pull/3

PR #266, opened 2015-07-03
NoSuchMethodException on bigendian system

santhosh-tekuri:
java.lang.NoSuchMethodException: org.msgpack.core.buffer.MessageBufferBE.<init>([B)
on MessageBuffer.java:146


xerial:
Thanks for the fix.

Issue #277, opened 2015-09-11
MessageUnpacker.unpackString returns invalid String


verandar:
```
import java.io.ByteArrayInputStream;
import java.io.ByteArrayOutputStream;
import java.io.IOException;
import java.io.InputStream;

import org.msgpack.core.MessagePacker;
import org.msgpack.core.MessageUnpacker;
import org.msgpack.core.buffer.InputStreamBufferInput;
import org.msgpack.core.buffer.OutputStreamBufferOutput;

// jdk-8u60, UTF-8 source code
public class UnpackStringTest {

    public static void main(String[] args) throws IOException {
        test(8188); // true
        test(8189); // false: (0.7.0-M6, 0.7.0-p9), true: (0.7.0.p7)
    }

    private static void test(int tmpSize) throws IOException {
        String expected = "あ";

        byte[] tmp = new byte[tmpSize];

        ByteArrayOutputStream out = new ByteArrayOutputStream();
        try (MessagePacker packer = new MessagePacker(new OutputStreamBufferOutput(out))) {
            packer.writePayload(tmp);
            packer.packString(expected);
        }

        InputStream in = new ByteArrayInputStream(out.toByteArray());
        try (MessageUnpacker unpacker = new MessageUnpacker(new InputStreamBufferInput(in))) {
            unpacker.readPayload(tmp);
            String actual = unpacker.unpackString();

            System.out.println(expected.equals(actual));
        }
    }
}
```

xerial:
At a first glance, you should call MessagePacker.flush() or close() to get the correct pack result.
And packer.writePayload must follow packer.packByteArrayHeader(tmp.length)

verandar:
Thanks.
MessagePacker.flush() or close() should be called by try-with-resources.
And I don't think packer.packByteArrayHeader() is a problem.

Here is another example.
I added packer.packBinaryHeader() and unpacker.unpackBinaryHeader().
And I changed tmpSize.

```
import java.io.ByteArrayInputStream;
import java.io.ByteArrayOutputStream;
import java.io.IOException;
import java.io.InputStream;

import org.msgpack.core.MessagePacker;
import org.msgpack.core.MessageUnpacker;
import org.msgpack.core.buffer.InputStreamBufferInput;
import org.msgpack.core.buffer.OutputStreamBufferOutput;

// jdk-8u60, UTF-8 source code
public class UnpackStringTest {

    public static void main(String[] args) throws IOException {
        test(8187); // true
[...]

PR #284, opened 2015-10-27
Take care of incomplete multi-bytes at the tail of buffer

komamitsu:
Fix https://github.com/msgpack/msgpack-java/issues/277

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
@@ -1025,10 +1025,10 @@ public class MessageUnpacker
                 decodeBuffer.clear();
                 StringBuilder sb = new StringBuilder();
 
-                boolean incompleteMultiBytes = false;
+                boolean hasIncompleteMultiBytes = false;
                 while (cursor < strLen) {
                     int readLen = Math.min(position < buffer.size() ? buffer.size() - position : buffer.size(), strLen - cursor);
-                    if (incompleteMultiBytes) {
+                    if (hasIncompleteMultiBytes) {
                         // Prepare enough buffer for decoding multi-bytes character right after running into incomplete one
                         readLen = Math.min(config.getStringDecoderBufferSize(), strLen - cursor);
                     }
@@ -1036,7 +1036,7 @@ public class MessageUnpacker
                         throw new EOFException();
                     }
 
-                    incompleteMultiBytes = false;
+                    hasIncompleteMultiBytes = false;
                     ByteBuffer bb = buffer.toByteBuffer(position, readLen);
                     int startPos = bb.position();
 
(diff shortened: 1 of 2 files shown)


komamitsu:
@xerial Can you review this pull request? Especially I'm not sure whether I should create another configuration parameter and use it here https://github.com/msgpack/msgpack-java/pull/284/files#diff-6a54557ce799041aa918fa8fa48bc78dR1033 instead of using `config.getStringDecoderBufferSize()`.

msgpack-core/src/test/scala/org/msgpack/core/MessageUnpackerTest.scala
```diff
@@ -663,5 +664,30 @@ class MessageUnpackerTest extends MessagePackSpec {
...
+        val unpacker = new MessageUnpacker(new InputStreamBufferInput(new ByteArrayInputStream(out.toByteArray)))
+        val len = unpacker.unpackBinaryHeader
+        unpacker.readPayload(len)
+        val got = unpacker.unpackString
+        unpacker.close
+
+        got shouldBe expected
+      }
+
+      Seq("あ", "aあ", "あa", "あいうえおかきくけこさしすせそ").foreach { s =>
```
xerial: Could you replace this into \uXXXX based strings, to avoid using Japanese characters within the code?

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
```diff
@@ -1025,13 +1025,20 @@ public String unpackString()
                 decodeBuffer.clear();
                 StringBuilder sb = new StringBuilder();
 
+                boolean incompleteMultiBytes = false;
```
xerial: `hasIncompleteMultiBytes` might be better naming

komamitsu:
@xerial Addressed your comments.

msgpack-core/src/test/scala/org/msgpack/core/MessageUnpackerTest.scala
```diff
@@ -663,5 +664,30 @@ class MessageUnpackerTest extends MessagePackSpec {
...
+        val bytes = new Array[Byte](strLen)
+        val out = new ByteArrayOutputStream
+
+        val packer = MessagePack.newDefaultPacker(out)
+        packer.packBinaryHeader(bytes.length)
+        packer.writePayload(bytes)
+        packer.packString(expected)
+        packer.close
+
+        val unpacker = new MessageUnpacker(new InputStreamBufferInput(new ByteArrayInputStream(out.toByteArray)))
```
xerial: FYI, you can use `MessagePack.newUnpacker(out.toByteArray)` here.

xerial:
Thanks for the fix! Merged.

Issue #292, opened 2015-11-15
MessagePacker#packString is slow when outputting to FileOutputStream


komamitsu:
I found MessagePacker#packString is slow when outputting to FileOutputStream.
```
    private static void packStrings(String label, String s, int count, OutputStream outputStream)
            throws IOException
    {
        System.out.println(label);
        MessagePacker messagePacker = MessagePack.newDefaultPacker(outputStream);
        long start = System.currentTimeMillis();
        for (int i = 0; i < count; i++) {
            messagePacker.packString(s);
        }
        messagePacker.close();
        System.out.println(System.currentTimeMillis() - start);
    }

    public static void main(String[] args)
            throws IOException
    {
        int count = 5000000;
        String s = "0123456789";
        packStrings("ByteArrayOutputStream", s, count, new ByteArrayOutputStream());
        packStrings("FileOutputStream", s, count, new FileOutputStream("/tmp/x.bin"));
        packStrings("FileOutputStream + BufferedOutputStream", s, count, new BufferedOutputStream(new FileOutputStream("/tmp/x.bin")));
    }
```

```
ByteArrayOutputStream
1056
FileOutputStream
19084
FileOutputStream + BufferedOutputStream
854
```

PR #293, opened 2015-11-15
Avoid to flush an extra buffer when the string isn't too large in Mes…

komamitsu:
…sagePack#packString

Fix https://github.com/msgpack/msgpack-java/issues/292

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
@@ -50,6 +50,7 @@ public class MessagePack
         private final int stringEncoderBufferSize;
         private final int stringDecoderBufferSize;
         private final int packerBufferSize;
+        private final int packerLargeStringThreshold;
         private final int packerRawDataCopyingThreshold;
 
         public Config(
@@ -61,6 +62,7 @@ public class MessagePack
                 int stringEncoderBufferSize,
                 int stringDecoderBufferSize,
                 int packerBufferSize,
+                int packerLargeStringLengthThreshold,
                 int packerRawDataCopyingThreshold)
         {
             checkArgument(packerBufferSize > 0, "packer buffer size must be larger than 0: " + packerBufferSize);
@@ -75,6 +77,7 @@ public class MessagePack
             this.stringEncoderBufferSize = stringEncoderBufferSize;
             this.stringDecoderBufferSize = stringDecoderBufferSize;
             this.packerBufferSize = packerBufferSize;
+            this.packerLargeStringThreshold = packerLargeStringLengthThreshold;
             this.packerRawDataCopyingThreshold = packerRawDataCopyingThreshold;
         }
 
@@ -133,6 +136,11 @@ public class MessagePack
             return packerBufferSize;
         }
 
+        public int getPackerLargeStringLengthThreshold()
+        {
+            return packerLargeStringThreshold;
+        }
+
         public int getPackerRawDataCopyingThreshold()
         {
             return packerRawDataCopyingThreshold;
@@ -154,6 +162,7 @@ public class MessagePack
         private int stringEncoderBufferSize = 8192;
         private int stringDecoderBufferSize = 8192;
         private int packerBufferSize = 8192;
+        private int packerLargeStringLengthThreshold = 512;
         private int packerRawDataCopyingThreshold = 512;
 
         public Config build()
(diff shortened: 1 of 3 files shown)


komamitsu:
@xerial Can you review this pull request? This won't completely resolve https://github.com/msgpack/msgpack-java/issues/292 because it works only when the string is not so large. But it works in most cases and mitigates this problem.

msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
```diff
@@ -442,6 +442,14 @@ public MessagePacker packDouble(double v)
         return this;
     }
 
+    private void packSmallString(String s)
+            throws IOException
+    {
+        byte[] bytes = s.getBytes("UTF-8");
```
xerial: This involves a look up of character encoding type table. Use `MessagePack.UTF8`

msgpack-core/src/test/scala/org/msgpack/core/MessagePackerTest.scala
```diff
@@ -235,6 +236,25 @@ class MessagePackerTest
...
       up1.close
     }
+
+    "pack a lot of String within expected time" in {
+      val count = 500000
+
+      def measureDuration(outputStream: java.io.OutputStream) : Long = {
+        val packer = MessagePack.newDefaultPacker(outputStream)
+        val start = Platform.currentTime
+        for (i <- 0 to count) {
```
xerial: Scala's for-comprehension has its-own performance bottleneck. Use
```
var i = 0
while(i < count) {
  ...
  i +=  1
}
``

File: msgpack-core/src/test/scala/org/msgpack/core/MessagePackerTest.scala [hunk 3]
```diff
@@ -235,6 +236,25 @@ class MessagePackerTest
...
+          packer.packString("0123456789ABCDEF")
+        }
+        packer.close
+        Platform.currentTime - start
+      }
+
+      val baseDuration = measureDuration(new ByteArrayOutputStream())
+      val (_, fileOutput) = createTempFileWithOutputStream
+      val targetDuration = measureDuration(fileOutput)
+      targetDuration shouldBe <= (baseDuration * 3)
```
xerial: The performance of these two blocks can be affected by the execution order because of JVM warm-up, JIT compilation, etc. And that is why I used `time` and `block` for performance comparison: https://github.com/msgpack/msgpack-java/blob/v07-develop/msgpack-core/src/test/scala/org/msgpack/core/MessageUnpackerTest.scala#L309

For example, 
```
val t = time("bench", repeat=10) {
  block("A") { ... }
  blcok("B") { ... }
}
```
It executes A, B, A, B, A, B, ...  You can reference the execution time average without min/max values with `t("A").averageWithoutMinMax`

msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
```diff
@@ -154,6 +162,7 @@ public int getPackerRawDataCopyingThreshold()
         private int stringEncoderBufferSize = 8192;
         private int stringDecoderBufferSize = 8192;
         private int packerBufferSize = 8192;
+        private int packerLargeStringLengthThreshold = 512;
```
xerial: This parameter doesn't tell how this threshold value will be used. How about `packerSmallStringOptimizationThreshold`? And also add a comment `this parameter is subject to change` so that we can fix the optimization in future.

xerial:
According to https://blogs.oracle.com/xuemingshen/entry/faster_new_string_bytes_cs, JVM improved the performance of `String.getBytes` for small strings. So this optimization makes sense.

komamitsu:
Thanks for the comments. I'll address them tomorrow.

msgpack-core/src/test/scala/org/msgpack/core/MessagePackerTest.scala
xerial: And simply printing the above time block displays benchmark result like this:
```
info(t)

-unpack performance	total:3.213 sec. , count:  100, avg:0.032 sec. , core avg:0.031 sec. , min:0.027 sec. , max:0.130 sec.
  -v6             	total:0.171 sec. , count:  100, avg:1.706 msec., core avg:1.666 msec., min:1.407 msec., max:5.866 msec.
  -v7             	total:0.117 sec. , count:  100, avg:1.172 msec., core avg:1.156 msec., min:0.948 msec., max:3.007 msec.
```

komamitsu:
@xerial Updated!

xerial:
Merged. Thanks!

PR #294, opened 2015-11-15
Simplify the config object

xerial:
This PR removes Config.getXXX. Although there is some advantage of having getters (replacing the internal field name, add a hook point in future, etc.), this config object is only used for storing configuration values, so I think reducing the code size has much advantage over these benefits.


xerial:
@komamitsu Could you review this PR?

komamitsu:
@xerial Looks good

PR #301, opened 2015-12-21
Add msgpack-json/jackson performance test

xerial:


komamitsu:
@xerial Thanks for the pull request.

But unfortunately, writing unit tests in Scala for msgpack-jackson may not be a good option for me since I'm not so good at Scala and I prefer to write test codes in Java for Java source code. Also, I want to see the variance of durations.

So I updated the original benchmark codes in https://github.com/msgpack/msgpack-java/pull/303. This benchmark shows small variances. If you're okay, I want to merge it instead.

xerial:
Closed in favor of #303

PR #303, opened 2015-12-23
Improve jackson benchmark in java

komamitsu:


msgpack-jackson/src/test/java/org/msgpack/jackson/dataformat/benchmark/Benchmarker.java
```diff
@@ -0,0 +1,95 @@
...
+            this.second = second;
+        }
+    }
+
+    public void run(int count, int warmupCount)
+            throws Exception
+    {
+        List<Tuple<String, double[]>> benchmarksResults = new ArrayList<Tuple<String, double[]>>(benchmarkableList.size());
+
+        for (Benchmarkable benchmark : benchmarkableList) {
```
xerial: The order of running programs also affects the performance results in JVM. Simply running System.gc() is not sufficient.

msgpack-jackson/src/test/java/org/msgpack/jackson/dataformat/benchmark/Benchmarker.java
```diff
@@ -0,0 +1,95 @@
...
+            double[] durations = new double[count];
+
+            for (int i = 0; i < count + warmupCount; i++) {
+                if (i >= warmupCount) {
+                    System.gc();
```
xerial: And also what do you expect by running this command? There is not guarantee that JVM does something by calling gc().

msgpack-jackson/src/test/java/org/msgpack/jackson/dataformat/benchmark/Benchmarker.java
```diff
@@ -0,0 +1,95 @@
...
+        for (Tuple<String, double[]> benchmarkResult : benchmarksResults) {
+            printStat(benchmarkResult.first, benchmarkResult.second);
+        }
+    }
+
+    private void printStat(String label, double[] values)
+    {
+        StandardDeviation standardDeviation = new StandardDeviation();
+        System.out.println(label + ":");
+        System.out.println(String.format("  mean : %8.3f", StatUtils.mean(values)));
```
xerial: The mean value including outliers (min, max) is not a good measure in evaluating the performance. We also need to add median or mean (average) value without containing outliers (e.g., by removing max and min, or a few percent of the top results).

msgpack-jackson/src/test/java/org/msgpack/jackson/dataformat/benchmark/Benchmarker.java
komamitsu: The benchmark has 4 times warmup runs for each one and it avoids the affect of running order, I think. BTW, how does your benchmark avoid the order of running programs?

msgpack-jackson/src/test/java/org/msgpack/jackson/dataformat/benchmark/Benchmarker.java
komamitsu: Actually, with calling the method, the variance of benchmark durations decreases. I should investigate the mechanism later.

msgpack-jackson/src/test/java/org/msgpack/jackson/dataformat/benchmark/Benchmarker.java
komamitsu: According to the result of profiling, this method accounts for 1% of the total duration. Also, we compare the performance of Jackson and MessagePack relatively with this benchmark even if this method affects a bit.

msgpack-jackson/src/test/java/org/msgpack/jackson/dataformat/benchmark/Benchmarker.java
xerial: For example, if you have two programs P1, P2,  it runs the program P1, P2, P1, P2,... to reduce the order effect.

msgpack-jackson/src/test/java/org/msgpack/jackson/dataformat/benchmark/Benchmarker.java
xerial: I just meant we should see median or average (which is not biased by too distant values).

msgpack-jackson/src/test/java/org/msgpack/jackson/dataformat/benchmark/Benchmarker.java
komamitsu: Okay. The original benchmark code (https://github.com/msgpack/msgpack-java/blob/v07-develop/msgpack-jackson/src/test/java/org/msgpack/jackson/dataformat/benchmark/MessagePackDataformatPojoBenchmarkTest.java#L98-L130) does the same thing, right?

msgpack-jackson/src/test/java/org/msgpack/jackson/dataformat/benchmark/Benchmarker.java
komamitsu: Sorry, I misread your comment. I'll take care of it

komamitsu:
@xerial Can you review it again? Thanks

xerial:
ok. It looks good.

PR #304, opened 2015-12-24
Add support for external buffer pools to unpacker

frsyuki:
This pull-request makes it possible that MessageBufferInput implementations can allocate memory from external buffer pools such as Netty (netty-io) PooledByteBufAllocator or Jetty ByteBufferPool.
This pull-request consists with following changes:

* Unpacker contains at most 1 MessageBuffer instance. With this change, MessageBufferInput implementations can ensure that a MessageBuffer instance returned at `next()` method becomes unnecessary (releasable) when the `next()` method is called again.
* String decoding code is updated. It includes slight optimization that instantiates a String instance without using streaming decoder if there is a complete buffer.
* New MessageInsufficientBufferException exception is defined to distinguish end of buffer from EOFException happened from the underlaying IO code. This is not backward compatible change.
* New MessageNeverUsedFormatException exception is defined to explicitly distinguish 0xC1 from other unexpected errors. This is backward compatible change because MessageNeverUsedFormatException extends MessageFormatException.

I observed slight performance improvement (~7%) with a simple benchmark test built in MessageUnpackerTest test code (unpack performance). But I don't expect large performance improvement with this change but may be able to expect by optimizing MessageBufferInput implementations.


xerial:
TravisCI test fails when using MessageBufferU (universal buffer). To investigate this, we need to run the tests with:
`./sbt test -J-Dmsgpack.universal-buffer=true`

PR #310, opened 2015-12-26
Add utility methods to pack/unpack to/from byte arrays efficiently

frsyuki:
This pull-request assumes #304 and #305, and provides utility methods to use msgpack-java more efficiently. Intention of this pull-request is that users use msgpack-java using intuitive APIs and those code become automatically the fastest code.

* Added MessageBufferPacker class and `MessagePack.newBufferPacker()` method to serialize objects into a byte array (or list of byte arrays). This is easier than ByteArrayOutputStream and a lot more efficient. Internally, this implements the optimization mentioned at #305.
* Added `MessagePack.newDefaultUnpacker(byte[], int, int)` 
* Size of both input and output buffers are configurable using MessagePackFactory API.


msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
```diff
@@ -37,406 +37,100 @@
...
-            return this;
-        }
-
-        public ConfigBuilder packerRawDataCopyingThreshold(int threshold)
-        {
-            this.packerRawDataCopyingThreshold = threshold;
-            return this;
-        }
-    }
+    private static MessagePackFactory defaultFactory = new MessagePackFactory();
```
komamitsu: Storing default factory class as a mutable static variable doesn't seem a good idea. For instance, if a dependent library uses MessagePack and internally sets a customized factory, it'll unexpectedly affect the application that uses MessagePack.

msgpack-core/src/main/java/org/msgpack/core/MessagePackFactory.java
```diff
@@ -0,0 +1,203 @@
...
+import java.io.OutputStream;
+import java.nio.channels.WritableByteChannel;
+import java.nio.channels.ReadableByteChannel;
+
+import java.nio.charset.CodingErrorAction;
+import static org.msgpack.core.Preconditions.checkArgument;
+
+public class MessagePackFactory
+{
+    private int packerSmallStringOptimizationThreshold = 512;
```
komamitsu: MessagePacker and MessageUnpacker have the same values and they can unexpectedly have different values in the future. How about adding a constant value in packer or unpacker class and making this class and MessagePacker/MessageUnpacker use the constant value?

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBufferOutput.java
```diff
@@ -17,30 +17,60 @@
...
+     * If the previously allocated buffer is not flushed yet, this next method should discard
+     * it without writing it.
      *
-     * @param bufferSize the buffer size to retrieve
+     * @param mimimumSize the mimium required buffer size to allocate
      * @return
      * @throws IOException
      */
-    public MessageBuffer next(int bufferSize)
+    public MessageBuffer next(int mimimumSize)
```
komamitsu: `minimumSize`

msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
```diff
@@ -191,12 +173,18 @@ public void close()
...
             throws IOException
     {
-        if (buffer == null || position + numBytesToWrite >= buffer.size()) {
-            flush();
-            buffer = out.next(Math.max(config.packerBufferSize, numBytesToWrite));
+        if (buffer == null) {
+            buffer = out.next(mimimumSize);
+        }
+        else if (position + mimimumSize >= buffer.size()) {
+            out.writeBuffer(position);
```
komamitsu: These codes are almost same as `flush()`. How about moving the both codes into a new method?

msgpack-core/src/main/java/org/msgpack/core/buffer/ArrayBufferOutput.java
```diff
@@ -0,0 +1,136 @@
...
+//
+package org.msgpack.core.buffer;
+
+import java.util.List;
+import java.util.ArrayList;
+
+/**
+ * MessageBufferOutput adapter that packs data into list of byte arrays.
+ */
+public class ArrayBufferOutput
```
komamitsu: How about `BufferListOutput`? `array` might sound like byte array.

msgpack-core/src/main/java/org/msgpack/core/MessagePackFactory.java
```diff
@@ -0,0 +1,203 @@
...
+
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.channels.WritableByteChannel;
+import java.nio.channels.ReadableByteChannel;
+
+import java.nio.charset.CodingErrorAction;
+import static org.msgpack.core.Preconditions.checkArgument;
+
+public class MessagePackFactory
```
komamitsu: Why don't you divide this class into `MessagePackerFactory` and `MessageUnpackerFactory` so that we can remove verbose prefixes such as `packerXXXXX` and `unpackerXXXXX`?

Also, I found msgpack-java 0.7 users already use jackson-dataformat-msgpack and they've written this code `new ObjectMapper(new MessagePackFactory())` in their source code. If we separate this class into `MessagePackerFactory` and `MessageUnpackerFactory`, we don't need to rename `org.msgpack.jackson.dataformat.MessagePackFactory` to `org.msgpack.jackson.dataformat.MessagePackFormatFactory` and the users don't need to change their source code.

msgpack-core/src/main/java/org/msgpack/core/buffer/ArrayBufferOutput.java
xerial: I think ArrayBufferOutput is fine since toByteArray() is available in this class. Having a list of MessageBuffer is just an implementation detail, so no need exists to reflect it to the class name.

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBufferOutput.java
xerial: fixed

msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
xerial: Thanks. Extracted the common code as flushBuffer()

msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
xerial: Good point. This setter for the default factory should be removed.

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
```diff
@@ -125,20 +132,43 @@
...
     }
 
-    /**
-     * Create an MessageUnpacker
-     *
-     * @param in
-     * @param config configuration
-     */
-    public MessageUnpacker(MessageBufferInput in, MessagePack.Config config)
+    public MessageUnpacker setAllowStringAsBinary(boolean enabled)
```
xerial: Do we need these configuration setters in MessageUnpacker? I think these method should be moved to the factory (or config) side.

msgpack-core/src/main/java/org/msgpack/core/MessagePackFactory.java
xerial: Instead of taking MessgePackFactory name from msgpack-jackson, introducing factory methods as MessagePack.Packer/UnpakerConfig.newXXX would be good for backward compatibility. 

I'll fix the code.

xerial:
@komamitsu I have finished the code review and applied your comments and minor modifications. Let me know if you finish the review. I'll merge this PR (by resolving conflicts with the v07-develop)

msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
```diff
@@ -659,72 +676,82 @@ else if (len < (1 << 16)) {
...
+     * @param src the data to add
+     * @param off the start offset in the data
+     * @param len the number of bytes to add
+     * @return this
+     * @throws IOException
+     */
+    public MessagePacker writePayload(byte[] src, int off, int len)
+            throws IOException
+    {
+        if (buffer.size() - position < len || len > 8192) {
```
komamitsu: We'd better create a constant variable and replace `8192` with it.

msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
```diff
@@ -442,14 +428,46 @@ public MessagePacker packDouble(double v)
...
+        if (cr.isError()) {
+            try {
+                cr.throwException();
+            }
+            catch (CharacterCodingException e) {
+                throw new MessageStringCodingException(e);
+            }
+        }
+        if (cr.isUnderflow() || cr.isOverflow()) {
+            return -1;
```
komamitsu: In `MessagePacker#packString()`, this return value `-1` is ignored. Why don't you throw `MessageStringCodingException`?

msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
```diff
@@ -464,77 +482,76 @@ public MessagePacker packString(String s)
...
-        encoder.reset();
-        while (in.hasRemaining()) {
-            try {
-                CoderResult cr = encoder.encode(in, encodeBuffer, true);
-
-                // Input data is insufficient
-                if (cr.isUnderflow()) {
-                    cr = encoder.flush(encodeBuffer);
+        else if (s.length() < (1 << 8)) {
+            // ensure capacity for 2-byte raw string header + the maximum string size (+ 1 byte for falback code)
```
komamitsu: `fallback`?

msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
```diff
@@ -464,77 +482,76 @@ public MessagePacker packString(String s)
...
+                }
+                else {
+                    if (written >= (1 << 32)) {
+                        // this must not happen because s.length() is less than 2^16 and (2^16) * UTF_8_MAX_CHAR_SIZE is less than 2^32
+                        throw new IllegalArgumentException("Unexpected UTF-8 encoder state");
+                    }
+                    // move 2 bytes backward to expand 3-byte header region to 5 bytes
+                    buffer.putBytes(position + 5,
+                            buffer.array(), buffer.arrayOffset() + position + 3, written);
+                    // write 3-byte header header
```
komamitsu: `header header` -> `header` ?

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
```diff
@@ -74,10 +71,19 @@
...
+    private final boolean allowReadingStringAsBinary;
+    private final boolean allowReadingBinaryAsString;
+    private final CodingErrorAction actionOnMalformedString;
+    private final CodingErrorAction actionOnUnmappableString;
+    private final int stringSizeLimit;
+    private final int stringDecoderBufferSize;
 
     private MessageBufferInput in;
 
+    private byte headByte = HEAD_BYTE_REQUIRED;
```
komamitsu: We can remove this variable and `HEAD_BYTE_REQUIRED`

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
```diff
@@ -169,195 +169,121 @@ public long getTotalReadBytes()
...
-     * MessageFormat instance. If false, {@link #getNextFormat()} will throw an EOFException.
+     * MessageFormat instance. If false, next {@link #getNextFormat()} call will throw an MessageInsufficientBufferException.
      *
      * @return true if this unpacker has more elements to read
      */
     public boolean hasNext()
             throws IOException
     {
-        return ensure(1);
+        while (buffer.size() <= position) {
```
komamitsu: It should be `if` instead of `while` here?

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
```diff
@@ -169,195 +169,121 @@ public long getTotalReadBytes()
...
-            int bufferSize = buffer.size();
-            position -= bufferSize;
-            totalReadBytes += bufferSize;
-            buffer = takeNextBuffer();
+        MessageBuffer next = in.next();
+        if (next == null) {
+            throw new MessageInsufficientBufferException();
         }
-        return buffer != null;
+        totalReadBytes += buffer.size();
```
komamitsu: Check if `buffer` is not null.

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
```diff
@@ -169,195 +169,121 @@ public long getTotalReadBytes()
...
-        // The buffer contains the data
-        if (position + byteSizeToRead <= buffer.size()) {
-            // OK
-            return true;
+        int remaining = buffer.size() - position;
+        if (remaining >= readLength) {
+            // When the data is contained inside the default buffer
+            nextReadPosition = position;
+            position += readLength;  // here assumes following buffer.getXxx never throws exception
+            return buffer; // Return the default buffer
```
komamitsu: [trivial] Why don't we return `buffer.slice(position, readLength)` so that we can remove `nextReadPosition`?

xerial: buffer.slice will create a new object and it will slow down the performance.

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
xerial: OK. I'll add an assertion. But there is no timing that the buffer will be null.

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
xerial: No. MessageBufferInput might return 0-size buffer.

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
xerial: ok.

msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
komamitsu: :+1:

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
```diff
@@ -586,7 +479,7 @@ public ImmutableValue unpackValue()
         MessageFormat mf = getNextFormat();
         switch (mf.getValueType()) {
             case NIL:
-                unpackNil();
+                readByte();
```
komamitsu: Non-`nil` value is allowed?

xerial: Nil is 1-byte format and this is already checked with mf.getValueType(). readByte() is used for forwarding the buffer position.

komamitsu: Ah, `mf.getValueType() == NIL` means the value is `nil`. Okay.

How about changing https://github.com/msgpack/msgpack-java/pull/310/files#diff-6a54557ce799041aa918fa8fa48bc78dR537 as well?

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
```diff
@@ -1006,107 +891,161 @@ public double unpackDouble()
...
+        if (decoder == null) {
+            decodeBuffer = CharBuffer.allocate(stringDecoderBufferSize);
+            decoder = MessagePack.UTF8.newDecoder()
+                    .onMalformedInput(actionOnMalformedString)
+                    .onUnmappableCharacter(actionOnUnmappableString);
+        }
+        else {
+            decoder.reset();
+        }
+        decodeStringBuffer = new StringBuilder();
```
komamitsu: `decodeStringBuffer.setLength(0)` can be a bit faster.

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
xerial: ok.

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
xerial: That's nice :)

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
```diff
@@ -1006,107 +891,161 @@ public double unpackDouble()
...
-                    }
-                    if (!ensure(readLen)) {
-                        throw new EOFException();
+        try {
+            int rawRemaining = len;
+            while (rawRemaining > 0) {
+                int bufferRemaining = buffer.size() - position;
+                if (bufferRemaining >= rawRemaining) {
+                    decodeStringBuffer.append(decodeStringFastPath(rawRemaining));
+                    rawRemaining = 0;
```
komamitsu: [trivial] Unnecessary code.

msgpack-core/src/main/java/org/msgpack/core/buffer/ChannelBufferInput.java
```diff
@@ -29,8 +29,7 @@
         implements MessageBufferInput
 {
     private ReadableByteChannel channel;
-    private boolean reachedEOF = false;
-    private final int bufferSize;
+    private final MessageBuffer m;
```
komamitsu: How about `buffer` or something instead of one character variable name?

komamitsu:
LGTM

frsyuki:
@xerial @komamitsu Thank you!

frsyuki:
@xerial about MessagePack.Code vs MessageFormat.Code: I wanted users to not aware of formats because most of users care about semantics (API) but don't want to care about syntax (formats). I first moved MessagePack.Code to MessageFormat.Code because MessagePack is the entry point (which is used by most of users) and users will need to care about methods/classes in it, and that's not expected to have Code there.

frsyuki:
@xerial about MessagePackFactory: I preferred to have a single FactoryClass to make it easy to configure MessagePack as a single library (maybe using a configuration file). I expected following example use case:

```java
MessagePackFactory factory = MyIoManagerConfigurator.loadMessagePackConfiguration("myio.config");
MessagePacker packer = factory.newPacker(...);
MessageUnpacker unpacker = factory.newUnpacker(...);
```

But as you mention, having 2 separated classes contributes to shorten parameter names. So, it was just a preference.

frsyuki:
@xerial about MessagePackFactory.inputBufferSize and MessagePackFactory.outputBufferSize. Sorry, my code had a bug where `newPacker(OutputStream)`, `newPacker(WritableChannel)`, `newUnpacker(InputStream)`, `newUnpacker(ReadableByteChannel)` should give inputBufferSize or outputBufferSize to constructors of OutputStreamBufferOutput, ChannelBufferOutput, InputStreamBufferInput or ChannelBufferInput. And this problem still exists in the latest code.

I think buffer size should be configurable because it's a trade-off of performance and memory consuption.

xerial:
@frsyuki Defining Code inside MessagePackFormat introduces unnecessary dependencies between MessagePacker/Unpacker to MessageFormat. So extracting Code class outside MessagePack might be better.

PR #312, opened 2015-12-27
Fix MessageFormat.Code#isFixedMap

komamitsu:
@frsyuki Can you review this fix? The bug has existed since v07-development, though.


frsyuki:
:+1:

Issue #313, opened 2015-12-30
[0.7] backward compatibility with previous serialization api (0.6)


marenzo:
I'm trying to upgrade from 0.6 to 0.7, and from what I saw the protocol itself is backward compatible.
We're having a lot of services out there which are using msgpack to communicate between them (RPC), and for doing this upgrade I must that the previous client (0.6) will know to deserialize 0.7 output, and vice-versa.

For now I'm having this two issues:
- Having the following code:
```java
if (len < (1 << 5)) {
    writeByte((byte) (FIXSTR_PREFIX | len));
}
else if (len < (1 << 8)) {
    writeByteAndByte(STR8, (byte) len);
}
else if (len < (1 << 16)) {
    writeByteAndShort(STR16, (short) len);
}
else {
    writeByteAndInt(STR32, len);
}
```
STR8 is not something that the previous client knows how to handle, and the fix I did locally was just commenting out ```else if (len < (1 << 8))```. This thing should be probably configurable, if you want to provide easier migration path.

- With the previous client, we serialized the values as Object array, that contained the fields the RPC needs to pass (may be primitives, String, Pojo's, and so on).
The way we did serializing with the previous client was:
```java
final ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
final Packer packer = msgPack.createPacker(outputStream);

final Object[] params;
if (requestParams == null) {
  params = new Object[0];
} else {
  params = requestParams;
}

packer.writeArrayBegin(params.length);
for (final Object param : params) {
  packer.write(param);
}
packer.writeArrayEnd();

return outputStream.toByteArray();
```
now, with the new jackson databind I'm doing:
```java
final ObjectMapper objectMapper = new ObjectMapper(new MessagePackFactory());
final Object[] params = new Object[]{ new MyPojo() }; // for example, may contain any other supported type / pojo
objectMapper.writeValueAsBytes(params);
```
and the byte array produced is not the same as the previous client, and it also contain all the fields of the pojo(?), which seems to miss one of the points of msgpack.
but if I'm also doing: ```objectMapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);```, the byte array not contains anymore all the field names, but still the produced byte array is not the same as with the previous client, which makes issues when trying to do de-serialize.

My questions are: is 0.7 should be backward compatible with 0.6 (in terms of consuming / producing same byte array)?
[...]

PR #314, opened 2015-12-30
changing str8 format to be configurable (to use it or not), and injecting Config object to MessagePackFactory

marenzo:
this pull request changes two things:
1. makes str8 format to be configurable (on or off), mainly for making migration to newer client easier (for people who aren't using latest 0.6 version, and want to be compatible with 0.7). See issue #313 
2. changing MessagePackFactory to have another c'tor which providing the ability of injecting configuration to MessagePacker, as today there's no way to do. For this change I had to drop ThreadLocal in MessagePackGenerator, but after checking carefully and doing concurrent tests, I believe there's no reason keeping it that way, and MessagePacker is not expensive to create.

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
@@ -214,9 +214,9 @@ public class MessagePack
     }
 
     /**
-     * Default configuration, which is visible only from classes in the core package.
+     * Default configuration
      */
-    static final Config DEFAULT_CONFIG = new ConfigBuilder().build();
+    public static final Config DEFAULT_CONFIG = new ConfigBuilder().build();
 
     /**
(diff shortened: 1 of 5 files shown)


xerial:
@marenzo 
Sorry for the delay of the response. 
We are now shifting to msgpack 0.8, which changes Config class into UnpackerConfig.

Basically your change looks good. Do you want to apply this patch for both 0.7 and 0.8 (with some modifications)?

marenzo:
@xerial 
Sure, I'll apply this change also to 0.8.
Do you want to marge this change to 0.7?

xerial:
We don't need to apply this change for msgpack 0.7.x.

Please re-create PR based on 0.8.
Thanks.

marenzo:
closing this PR in favour of new one (#356)

PR #320, opened 2016-01-05
Make it possible to make ArrayBufferInput empty

frsyuki:
This change adds support to give null to ArrayBufferInput.reset method.
This lets applications control lifecycle of MessageBuffer (such as
releasing it to a buffer pool) before calling ArrayBufferInput.close().

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/buffer/ArrayBufferInput.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/buffer/ArrayBufferInput.java
@@ -26,11 +26,17 @@ public class ArrayBufferInput
         implements MessageBufferInput
 {
     private MessageBuffer buffer;
-    private boolean isRead = false;
+    private boolean isEmpty;
 
     public ArrayBufferInput(MessageBuffer buf)
     {
-        this.buffer = checkNotNull(buf, "input buffer is null");
+        this.buffer = buf;
+        if (buf == null) {
+            isEmpty = true;
+        }
+        else {
+            isEmpty = false;
+        }
     }
 
     public ArrayBufferInput(byte[] arr)
@@ -54,10 +60,10 @@ public class ArrayBufferInput
         MessageBuffer old = this.buffer;
         this.buffer = buf;
         if (buf == null) {
-            this.isRead = true;
+            isEmpty = true;
         }
         else {
-            this.isRead = false;
+            isEmpty = false;
         }
         return old;
     }
@@ -76,10 +82,10 @@ public class ArrayBufferInput
     public MessageBuffer next()
             throws IOException
     {
-        if (isRead) {
+        if (isEmpty) {
             return null;
         }
-        isRead = true;
+        isEmpty = true;
         return buffer;
     }
 
@@ -88,6 +94,6 @@ public class ArrayBufferInput
             throws IOException
     {
         buffer = null;
-        isRead = true;
+        isEmpty = true;
     }
 }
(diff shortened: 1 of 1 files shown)


xerial:
Allowing null looks weird. Does it really need to pass null? Having `clean()` method to release the internal buffer is generally better.

frsyuki:
I expect following application code:

```java
public class SimpleBufferPool {
    private Deque<MessageBuffer> pool;

    public synchronized MessageBuffer allocate(int minimumRequiredSize) {
        MessagePack buffer = pool.peekFirst();
        if (buffer != null && buffer.size() >= minimumRequiredSize) {
            return pool.removeFirst();
        }
        else {
            return MessageBuffer.allocate(Math.max(32*1024, minimumRequiredSize));
        }
    }

    public synchronized void release(MessageBuffer buffer) {
        if (buffer != null) {
            pool.offerLast(buffer);
        }
    }
}

public class MySocketHandler {
    private final DataInputStream socket;
    private final MessagePack.UnpackerConfig unpackerConfig;
    private final SimpleBufferPool pool;

    public MySocketHandler(DataInputStream socket) {
        this(socket, new MessagePack.UnpackerConfig());
    }

    public MySocketHandler(DataInputStream socket, MessagePacker.UnpackerConfig unpackerConfig) {
        this.socket = socket;
        this.unpackerConfig = unpackerConfig;
        this.pool = create pool ...();
    }

    public void main() {
        ArrayBufferInput in = new ArrayBufferInput(null);
        try (messageUnpacker unpacker = unpackerConfig.newUnpacker(in)) {
            while (true) {
                int messageSize = socket.readInt();
                MessageBuffer buffer = pool.allocate(messageSize)
                in.reset(buffer);
                socket.readFully(buffer.array(), buffer.arrayOffset(), buffer.size());
                try {
                    unpacker.reset(in);
                    final Value value = unpacker.unpackValue();
                    new Thread(() -> asyncProcessValue(value));
                }
                finally {
                    pool.release(in.reset(null));
                }
            }
        }
        catch (IOException ex) {
            ...
        }
    }

    public void asyncProcessValue(Value value) {
        ...
    }
}
```

Or, as a replacement of SimpleBufferPool:

```java
public class AdvancedBufferPool {
    private io.netty.buffer.PooledByteBufAllocator pool;
    private IdentityHashMap<MessageBuffer, io.netty.buffer.ByteBuf> map;

    public synchronized MessageBuffer allocate(int minimumRequiredSize) {
        ByteBuf buf = pool.buffer(Math.max(32*1024, minimumRequiredSize));
        buffer = MessageBuffer.wrap(buf.array(), buf.arrayOffset(), buf.capacity());
        map.put(buffer, buf);
        return buffer;
    }

    public synchronized void release(MessageBuffer buffer) {
        if (buffer != null) {
            ByteBuf buf = map.remove(buffer);
            if (buf == null) {
                throw new NullPointerException("Detected double-free");
            }
            buf.release();
        }
    }
}
```

This code assumes that `MySocketHandler.main` can create an unpacker with an empty ArrayBufferInput without allocating buffer so that it doesn't hold unnecessary memory until it gets the first 4 bytes from a socket.

frsyuki:
If null seems unacceptable, another idea is to have MessageBuffer.empty() that returns a consistent immutable empty MessageBuffer instance. Then above can be:

```java
public class SimpleBufferPool {
    ...

    public synchronized void release(MessageBuffer buffer) {
        if (buffer != MessageBuffer.empty()) {
            pool.offerLast(buffer);
        }
    }
}

public class MySocketHandler {
    ...

    public void main() {
        ArrayBufferInput in = new ArrayBufferInput(MessageBuffer.empty());
        try (messageUnpacker unpacker = unpackerConfig.newUnpacker(in)) {
            while (true) {
                ...
                try {
                    ...
                }
                finally {
                    pool.release(in.reset(MessageBuffer.empty()));
                }
            }
        }
        catch (IOException ex) {
            ...
        }
    }

    ...
}
```

xerial:
OK. I got the application. Then allowing null is OK since providing MessageBuffer.empty() will be a little bit tricky since we may need to manage MessageBufferU, MessageBufferBE.empty(), etc.

PR #321, opened 2016-01-05
Make constructors protected to ensure the single entry point

frsyuki:
Makes constructors of MessagePacker and MessageUnpacker protected so
that all users use factory methods of MessagePack interface. This
removes duplication of API and make it easy to understand.

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/MessageBufferPacker.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessageBufferPacker.java
@@ -28,12 +28,12 @@ import java.util.List;
 public class MessageBufferPacker
         extends MessagePacker
 {
-    public MessageBufferPacker(MessagePack.PackerConfig config)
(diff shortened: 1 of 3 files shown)


frsyuki:
This follows-up #310.
@xerial I think having the single entry point makes it easier to understand/explain/document the API.

frsyuki:
And I think, usually, such entry points are called Factory rather than Config. How do you think about using `MessagePack.PackerFactory` and `MessagePack.UnpackerFactory` instead of `MessagePack.PackerConfig` and `MessagePack.UnpackerConfig`? This is minor topic relatively.

xerial:
@frsyuki 
Using protected is ok. 

I talked with @komamitsu about class naming. Using xxxFactory both for MessagePacker/Unpacker and Jackson binding is confusing. And I also agree that Config does not usually have a factory method.

How about providing the following method so that user can notice config object is necessary to configure packer and unpacker?  
MessagePack.newPacker(PackerConfig, out), newUnpacker(UnpackerConfig, out)

PR #324, opened 2016-01-05
Make copy and reference clear at ValueFactory

frsyuki:
ValueFactory .newBinary, .newString, .newArray and .newMap take a
mutable argument. Some of them used them as a reference and some of them
copied them. This pull-request makes it clear that they copy the
argument by default and try to omit the copy if an extra argument is
set to true.

This change also includes:

* newMap(Value... kvs) method which was previously newMap(Value[] kvs)
* MessageUnpacker.unpackValue removes a duplicated copy of new byte[]
  (string and binary) or Value[] (array and map).

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/value/ValueFactory.java
+++ b/msgpack-core/src/main/java/org/msgpack/value/ValueFactory.java
@@ -87,12 +87,32 @@ public final class ValueFactory
 
     public static ImmutableBinaryValue newBinary(byte[] b)
     {
-        return new ImmutableBinaryValueImpl(b);
+        return newBinary(b, false);
+    }
+
+    public static ImmutableBinaryValue newBinary(byte[] b, boolean ref)
+    {
+        if (ref) {
+            return new ImmutableBinaryValueImpl(b);
+        }
+        else {
+            return new ImmutableBinaryValueImpl(Arrays.copyOf(b, b.length));
+        }
     }
 
     public static ImmutableBinaryValue newBinary(byte[] b, int off, int len)
     {
-        return new ImmutableBinaryValueImpl(Arrays.copyOfRange(b, off, len));
+        return newBinary(b, off, len, false);
+    }
+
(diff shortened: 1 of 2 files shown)


frsyuki:
@xerial Another possible design is adding a new method (such as newBinaryRef, newArrayRef) instead of adding an extra argument. With this way, users will expect that -Ref methods will always use reference. But in fact, they use reference only if `off == 0 && len == b.length` because ImmutableBinaryValue and ImmutableStringValue (in terms of implementation, AbstractImmutableRawValue) don't accept offset and length. So AbstractImmutableRawValue also needs more code. An ideas?

msgpack-core/src/main/java/org/msgpack/value/ValueFactory.java
```diff
@@ -87,12 +87,32 @@ public static ImmutableFloatValue newFloat(double v)
 
     public static ImmutableBinaryValue newBinary(byte[] b)
     {
-        return new ImmutableBinaryValueImpl(b);
+        return newBinary(b, false);
+    }
+
+    public static ImmutableBinaryValue newBinary(byte[] b, boolean ref)
```
xerial: The meaning of `ref` is a little bit ambiguous. How about using `boolean createCopy`?

frsyuki: The "correct" name is `useReferenceIfPossible` but it seems too long. How about `boolean omitCopy` instead?
I think inverting true/false is confusing because using true as the default seems unexpected.

xerial: `omitCopy` is fine.

frsyuki: I changed it to `omitCopy`.

xerial:
ValueFactory is just a collection of helper methods and users simply need to use newXXX methods without `omitCopy` flag. In this sense, MessageUnpacker is the only user who cares about the performance when creating Value objects. I think simply having the omitCopy flag is sufficient.

PR #325, opened 2016-01-06
Support method-chain syntax to create packer and unpacker with immutable config

frsyuki:
* This change allows applications to use method-chain syntax to create packer and unpacker with config. For example,

```java
MessageUnpacker unpacker = new UnpackerConfig()
    .withStringDecoderBufferSize(16 * 1024)
    .withActionOnMalformedString(CodingErrorAction.REPORT)
    .withActionOnUnmappableString(CodingErrorAction.REPORT)
    .newUnpacker(packedData);
```

* This change makes {Packer,Unpacker}Config immutable. This makes it safe to share a config object.

* This change makes buffer size of InputStreamBufferInput, OutputStreamBufferOutput, and ChannelBuffer{Input,Output} configurable. Configurable buffer size is better because it is trade-off of performance and memory consumption.

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
@@ -235,17 +235,9 @@ public class MessagePack
      */
     public static class PackerConfig
     {
-        /**
-         * Use String.getBytes() for converting Java Strings that are smaller than this threshold into UTF8.
-         * Note that this parameter is subject to change.
-         */
-        public int smallStringOptimizationThreshold = 512;
+        private int smallStringOptimizationThreshold = 512;
 
-        /**
-         * When the next payload size exceeds this threshold, MessagePacker will call MessageBufferOutput.flush() before
-         * packing the data.
-         */
-        public int bufferFlushThreshold = 8192;
+        private int bufferFlushThreshold = 8192;
 
         /**
          * Create a packer that outputs the packed data to a given output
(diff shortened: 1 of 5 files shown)


msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
```diff
@@ -289,42 +312,112 @@ public MessageBufferPacker newBufferPacker()
...
+        }
+
+        @Override
+        public UnpackerConfig clone()
+        {
+            return new UnpackerConfig(this);
+        }
+
+        @Override
+        public boolean equals(Object obj)
```
xerial: Please add a test code to check the equality of config objects

msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
```diff
@@ -289,42 +312,112 @@ public MessageBufferPacker newBufferPacker()
...
+        }
+    }
+
+    /**
+     * MessageUnpacker configuration.
+     */
+    public static class UnpackerConfig
+            implements Cloneable
+    {
+        private boolean allowReadingStringAsBinary = true;
```
xerial: Now that Packer/UnpackerConfig objects are immutable, we can make them public final to remove redundant getter methods. What do you think?

msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
frsyuki: added

msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
frsyuki: Using getter is safer here because {Packer,Unpacker}Config are public API and change of the API means backward incompatibility. Exposing internal data structure seems negative for the future development.

xerial: Then we may need to use @deprecated annotation for the future config change. OK.

PR #329, opened 2016-01-06
MessageBufferPacker.toByteArray flushes packer's internal buffer automatically

frsyuki:
otherwise applications need to call flush() for sure, and we often miss it...


frsyuki:
@xerial can you also check the test code? I'm not familiar with Scala.

xerial:
The test code looks good. I just slightly modified the coding style.

PR #334, opened 2016-01-13
Extension type is not a raw type

frsyuki:
ExtensionValue doesn't extend RawValue.
Only BinaryValue and StringValue extend RawValue.


xerial:
Looks good.

PR #336, opened 2016-01-13
ValueFactory.MapBuilder is preferred to preserve value order

frsyuki:
ImmutableMapValueImpl preserves order of values. This is preferred
behavior because it's preffered if created map objects preserve order of
values when an application forwards a received (or deserialized) objects
into another application without modifying it. This pull-request applies
this preference to MapBuilder.

Performance impact should be small because MapBuilder API is a minor
API.


xerial:
I also hit this problem when using msgpack v8 for plazmadb-mpcfile. Thanks for the fix.

PR #341, opened 2016-01-30
Copy old data before fetching new data, to avoid overwriting it

MikePlayle:
Hi!

I encountered a bug while trying to decode integers that cross buffer boundaries. The problem is that MessageUnpacker.prepareNumberBuffer() calls getNextBuffer() before it has copied the data out of the old buffer, but InputStreamBufferInput.next() reuses the existing byte array. This means that by the time it returns, the remaining bytes from the old buffer have been overwritten with fresh data.

This pull request restructures prepareNumberBuffer() so that it copies the remaining bytes before asking for a new buffer. It fixes my problem and still passes its own test suite.

(This is my first attempt to use GitHub, so please be patient if I've got anything wrong!)

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
@@ -208,32 +208,36 @@ public class MessageUnpacker
             position += readLength;  // here assumes following buffer.getXxx never throws exception
             return buffer; // Return the default buffer
         }
+        else if(remaining == 0) {
+            buffer = getNextBuffer();
+            position = readLength;
+            nextReadPosition = 0;
+            return buffer;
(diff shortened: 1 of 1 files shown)


xerial:
Thanks for reporting. I'll check it later.

xerial:
@komamitsu The fix looks good. Could you merge this and add your test case in #346 (by making it more precise to read the buffer boundary)?

xerial:
@komamitsu Here is an example to write a test at buffer boundary: https://github.com/msgpack/msgpack-java/blob/develop/msgpack-core/src/test/scala/org/msgpack/core/MessageUnpackerTest.scala#L265

xerial:
We added a test case to reproduce this issue and merged your fix. Thanks for your contribution.

PR #346, opened 2016-02-20
Fix bug MessageUnpacker#prepareNumberBuffer reads updated buffer

komamitsu:
I ran into a bug that MessageUnpacker#unpack(Int|Long|Float|Double) occasionally returns the wrong value when using InputStreamBufferInput. And I found MessageUnpacker#prepareNumberBuffer reads the tail of ` buffer` implicitly updated by InputStreamBufferInput#next with original position when the remaining size of `buffer` is shorter than read length.

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
@@ -211,15 +211,16 @@ public class MessageUnpacker
         else {
             // When the default buffer doesn't contain the whole length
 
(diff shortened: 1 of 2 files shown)


komamitsu:
@frsyuki or @xerial Can you review this pull request?

xerial:
@komamitsu Is it related to #341?

komamitsu:
Yeah. I didn't notice it. The fix is almost same as mine.

xerial:
Let me close this PR since this is fixed in #348

PR #350, opened 2016-02-23
Fix msgpack core usage example

xerial:
@naritta 
Could you check this fix? Just changed to use MessageBufferPacker in the example codes.
Thanks.


naritta:
Looks good and it works well.
Thanks.

PR #356, opened 2016-04-06
changing str8 format to be configurable (to use it or not), and injecting Config object to MessagePackFactory

marenzo:
a new pull request after the discussion in the previous one (#314, #313).
this time the patch is made for msgpack 0.8
the changes are same as in the previous pull request (plus changes of the new configuration objects).

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
@@ -38,6 +38,12 @@ public class MessagePack
     public static final Charset UTF8 = Charset.forName("UTF-8");
 
     /**
+     * Default packer/unpacker configurations
+     */
+    public static final PackerConfig DEFAULT_PACKER_CONFIG = new PackerConfig();
+    public static final UnpackerConfig DEFAULT_UNPACKER_CONFIG = new UnpackerConfig();
+
+    /**
      * The prefix code set of MessagePack. See also https://github.com/msgpack/msgpack/blob/master/spec.md for details.
      */
     public static final class Code
@@ -138,7 +144,7 @@ public class MessagePack
      */
     public static MessagePacker newDefaultPacker(MessageBufferOutput out)
     {
-        return new PackerConfig().newPacker(out);
+        return DEFAULT_PACKER_CONFIG.newPacker(out);
     }
 
     /**
@@ -149,7 +155,7 @@ public class MessagePack
      */
     public static MessagePacker newDefaultPacker(OutputStream out)
     {
-        return new PackerConfig().newPacker(out);
+        return DEFAULT_PACKER_CONFIG.newPacker(out);
     }
 
     /**
@@ -160,7 +166,7 @@ public class MessagePack
      */
     public static MessagePacker newDefaultPacker(WritableByteChannel channel)
     {
-        return new PackerConfig().newPacker(channel);
+        return DEFAULT_PACKER_CONFIG.newPacker(channel);
     }
 
     /**
@@ -170,7 +176,7 @@ public class MessagePack
      */
     public static MessageBufferPacker newDefaultBufferPacker()
     {
-        return new PackerConfig().newBufferPacker();
+        return DEFAULT_PACKER_CONFIG.newBufferPacker();
     }
 
     /**
@@ -181,7 +187,7 @@ public class MessagePack
      */
     public static MessageUnpacker newDefaultUnpacker(MessageBufferInput in)
     {
-        return new UnpackerConfig().newUnpacker(in);
+        return DEFAULT_UNPACKER_CONFIG.newUnpacker(in);
     }
 
     /**
@@ -192,7 +198,7 @@ public class MessagePack
      */
     public static MessageUnpacker newDefaultUnpacker(InputStream in)
     {
-        return new UnpackerConfig().newUnpacker(in);
+        return DEFAULT_UNPACKER_CONFIG.newUnpacker(in);
     }
 
     /**
@@ -203,7 +209,7 @@ public class MessagePack
      */
     public static MessageUnpacker newDefaultUnpacker(ReadableByteChannel channel)
     {
-        return new UnpackerConfig().newUnpacker(channel);
+        return DEFAULT_UNPACKER_CONFIG.newUnpacker(channel);
     }
 
     /**
@@ -214,7 +220,7 @@ public class MessagePack
      */
     public static MessageUnpacker newDefaultUnpacker(byte[] contents)
     {
-        return new UnpackerConfig().newUnpacker(contents);
+        return DEFAULT_UNPACKER_CONFIG.newUnpacker(contents);
     }
 
     /**
@@ -227,7 +233,7 @@ public class MessagePack
      */
     public static MessageUnpacker newDefaultUnpacker(byte[] contents, int offset, int length)
     {
-        return new UnpackerConfig().newUnpacker(contents, offset, length);
+        return DEFAULT_UNPACKER_CONFIG.newUnpacker(contents, offset, length);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
```diff
@@ -38,6 +38,12 @@
     public static final Charset UTF8 = Charset.forName("UTF-8");
 
     /**
+     * Default packer/unpacker configurations
+     */
+    public static final PackerConfig DEFAULT_PACKER_CONFIG = new PackerConfig();
+    public static final UnpackerConfig DEFAULT_UNPACKER_CONFIG = new UnpackerConfig();
```
xerial: Make these variable scopes to package by removing `public` since these are only for internal use.

msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackGenerator.java
```diff
@@ -38,7 +38,7 @@
         extends GeneratorBase
 {
     private static final Charset DEFAULT_CHARSET = Charset.forName("UTF-8");
-    private static ThreadLocal<MessagePacker> messagePackersHolder = new ThreadLocal<MessagePacker>();
```
xerial: @komamitsu Could you tell me why you needed to use ThreadLoacl for storing MessagePacker?

msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackFactory.java
```diff
@@ -35,11 +36,23 @@
...
     private static final long serialVersionUID = 2578263992015504347L;
 
+    private final MessagePack.PackerConfig packerConfig;
+
+    public MessagePackFactory()
+    {
+        this(MessagePack.DEFAULT_PACKER_CONFIG);
+    }
+
+    public MessagePackFactory(final MessagePack.PackerConfig packerConfig)
```
xerial: Remove `final`. We are not enforcing this coding style for such a simple code.

msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackGenerator.java
```diff
@@ -95,11 +95,10 @@ void addKey(String key)
         }
     }
 
-    public MessagePackGenerator(int features, ObjectCodec codec, OutputStream out)
+    public MessagePackGenerator(int features, ObjectCodec codec, OutputStream out, final MessagePack.PackerConfig packerConfig)
```
xerial: Remove `final` here. The same reason as above.

msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackGenerator.java
```diff
@@ -109,14 +108,7 @@ public MessagePackGenerator(int features, ObjectCodec codec, OutputStream out)
         }
         messageBufferOutputHolder.set(messageBufferOutput);
 
-        if (messagePacker == null) {
-            messagePacker = MessagePack.newDefaultPacker(messageBufferOutput);
-        }
-        else {
-            messagePacker.reset(messageBufferOutput);
```
xerial: @marenzo The use of `reset()` here is for the performance. If we create a new MessagePacker instance every time, it will slow down the deserialization process.

msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackFactory.java
marenzo: fixed. my IDE settings puts final everywhere automatically :)

msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackGenerator.java
marenzo: fixed

msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
marenzo: I'm using those constants in MessagePackFactory instead of creating default packer config each time. We can either leave it public, or create another constant in MessagePackFactory to hold single instance. WDYT?

msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackGenerator.java
marenzo: I don't see how it can slow down the serialization process, as the creation of MessagePacker basically does nothing - just sets some fields. The thing you're right here is that I'm creating now MessagePacker for each serialization process - which means more garbage to remove later.
But because I want to provide the ability of passing config for each MessagePackFactory, I just can't hold shared state because else this feature just won't be possible.
Do you think there's better solution?

xerial: Right. The actual cost here is just an object instantiation cost. And it could be negligible if we serialize ~ 10,000 objects. 

@komamitsu What do you think?

msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
xerial: Ah I got it. MessagePackFactory is in another package scope. OK. Keep these definitions.

msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
```diff
@@ -360,6 +368,23 @@ public int getBufferSize()
         {
             return bufferSize;
         }
+
+        /**
+         * Disable str8 format when needed backward compatibility between
+         * different msgpack serializer versions.
+         * default true (str8 supported enabled)
+         */
+        public PackerConfig withStr8FormatSupport(boolean str8FormatSupport)
```
xerial: Fix PackerConfig.equals() method as well to involve this parameter.

marenzo: fixed and added the field to the copy c'tor after the latest fix.

xerial:
Overall looks good to me. @komamitsu could you answer above my comments?

msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackGenerator.java
komamitsu: MessagePackGenerator is created each time when ObjectMapper#writeValue is called. So MessagePackGenerator stored MessagePacker in ThreadLocal to reduce the cost of instantiation.

msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
```diff
@@ -267,7 +276,8 @@ public boolean equals(Object obj)
             PackerConfig o = (PackerConfig) obj;
             return this.smallStringOptimizationThreshold == o.smallStringOptimizationThreshold
                     && this.bufferFlushThreshold == o.bufferFlushThreshold
-                    && this.bufferSize == o.bufferSize;
+                    && this.bufferSize == o.bufferSize
```
komamitsu: @xerial We'd better override `hashCode` when overriding `equals`. Let's do that later

msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackGenerator.java
komamitsu: I benchmarked this pull request and saw that removing the caching of MessagePacker doesn't affect the performance while the caching of OutputStreamBufferOutput effects significantly. So the change looks good.

komamitsu:
LGTM as well. Thanks! @marenzo

xerial:
Merged. But I noticed packString also uses STR8 here:
https://github.com/msgpack/msgpack-java/blob/ca723f87ac667c5430ee1cc383555f3790cb08b7/msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java#L503-L503

I'll add a test code and fix this.

marenzo:
Thanks!

xerial:
Added a fix for packString #359

PR #358, opened 2016-04-07
adding helper for serialization format of object mapper to be array

marenzo:
related to the tip that @komamitsu gave in #313 issue, about adding `@JsonFormat(shape=JsonFormat.Shape.ARRAY)` annotation to made a pojo being serialized as array. The issue with that is most of the times you have tons of pojo's around, and you can't enforce all of them to have this annotation, so I've made some global helper to enforce this serialization format on all of the objects serialized by object mapper.

this helper also was made because jackson doesn't provide any programmatic way to do such thing.

### DIFF
--- a/msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/JsonArrayFormat.java
+++ b/msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/JsonArrayFormat.java
@@ -12,8 +12,6 @@ import static com.fasterxml.jackson.annotation.JsonFormat.Shape.ARRAY;
  * way.
  *
  * This also provides same behavior as msgpack-java 0.6.x serialization api.
- *
- * @author marenzo
  */
 public class JsonArrayFormat extends JacksonAnnotationIntrospector {
 
--- a/msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/JsonArrayFormat.java
+++ b/msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/JsonArrayFormat.java
@@ -1,0 +1,33 @@
+package org.msgpack.jackson.dataformat;
+
+import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.databind.introspect.Annotated;
+import com.fasterxml.jackson.databind.introspect.JacksonAnnotationIntrospector;
+
+import static com.fasterxml.jackson.annotation.JsonFormat.Shape.ARRAY;
+
+/**
+ * Provides the ability of serializing POJOs without their schema.
+ * Similar to @JsonFormat annotation with JsonFormat.Shape.ARRAY, but in a programmatic
+ * way.
+ *
+ * This also provides same behavior as msgpack-java 0.6.x serialization api.
+ *
+ * @author marenzo
+ */
+public class JsonArrayFormat extends JacksonAnnotationIntrospector {
+
+  private final static JsonFormat.Value ARRAY_FORMAT = new JsonFormat.Value().withShape(ARRAY);
+
+  @Override
+  public JsonFormat.Value findFormat(Annotated ann)
+  {
+    // If the entity contains JsonFormat annotation, give it higher priority.
+    JsonFormat.Value precedenceFormat = super.findFormat(ann);
+    if (precedenceFormat != null) {
+      return precedenceFormat;
+    }
+
+    return ARRAY_FORMAT;
+  }
+}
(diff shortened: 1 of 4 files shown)


xerial:
Just curiosity. How do you ensure the order of elements in an array when serializing objects?

marenzo:
good question, checking now Jackson's code to see what guarantees it offers (if any).
how did msgpack 0.6 worked in that matter? AFAIK, javassist doesn't provided guarantees about fields order

marenzo:
as far as I see, ObjectMapper uses java reflection api to receive pojo's fields (POJOPropertiesCollector calling AnnotatedClass#resolveFields, which eventually uses Class#getDeclaredFields).
the JVM doesn't provide any guarantees about fields order, although it seems that the deceleration order  
is being saved (can test more this statement).

xerial:
AFAK there is no guarantee of the field order returned by reflection. msgpack v6 has `index` annotation to ensure this. 

If the fields are inherited from parent/ancestor classes, I guess the field order will be more complicated.

marenzo:
Jackson have annotation called 'JsonPropertyOrder', which gives you the ability to define by yourself the fields order. So If someone would like to define manually the order, it's possible.

Basically I'll use this utility, as I'll need my pojos to be serialized without their scheme, so I thought this will be also common use for people who want to make a migration, or more of that, for people who want to keep serializing without scheme.

msgpack-jackson/src/test/java/org/msgpack/jackson/dataformat/MessagePackDataformatForPojoTest.java
```diff
@@ -99,4 +105,18 @@ public void testChangingPropertyNames()
...
+            throws IOException
+    {
+        ObjectMapper objectMapper = new ObjectMapper(factory); // to not affect shared objectMapper state
+        UsingCustomConstructorPojo orig = new UsingCustomConstructorPojo("komamitsu", 55);
+        objectMapper.setAnnotationIntrospector(new JsonArrayFormat());
+        byte[] bytes = objectMapper.writeValueAsBytes(orig);
+        String scheme = new String(bytes, Charset.forName("UTF-8"));
+        assertThat(scheme, not(containsString("name")));
+        UsingCustomConstructorPojo value = objectMapper.readValue(bytes, UsingCustomConstructorPojo.class);
+        assertEquals("komamitsu", value.name);
```
xerial: Do we need to add a test for '55' as well?

msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/JsonArrayFormat.java
```diff
@@ -0,0 +1,33 @@
...
+import static com.fasterxml.jackson.annotation.JsonFormat.Shape.ARRAY;
+
+/**
+ * Provides the ability of serializing POJOs without their schema.
+ * Similar to @JsonFormat annotation with JsonFormat.Shape.ARRAY, but in a programmatic
+ * way.
+ *
+ * This also provides same behavior as msgpack-java 0.6.x serialization api.
+ *
+ * @author marenzo
```
xerial: Please remove this `@author` tag. msgpack-java should be owned by public community :)

msgpack-jackson/src/test/java/org/msgpack/jackson/dataformat/MessagePackDataformatForPojoTest.java
marenzo: you're right, added

msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/JsonArrayFormat.java
marenzo: :) removed

xerial:
@marenzo FYI. You can run code style check with `./sbt jcheckStyle`

marenzo:
@xerial yep, sorry for the multiple commits. fixed

xerial:
@marenzo One more thing. I think we should add a test to confirm that the generated data is actually has an array type.

marenzo:
@xerial changing the test to use NestedListPojo would be fine, or we want some new pojo that will contains some primitives and data structures (list, map, set)?

xerial:
@marenzo It's up to you. Simply we need to make sure that we can produce array type data.

marenzo:
@xerial done. also made some small code style fixes in the test if you don't mind :)

xerial:
Nice work! I'll merge this and release a new version.

marenzo:
Cool, thanks!

xerial:
Released 0.8.5, which also includes the configuration of STR8 format. This will be available soon from the Maven central.

starthal:
Do I need to use Jackson to be able to pack a custom class in the array format? Can I just annotate with `@Message` and `@index` and expect the fields to be (de)serialized in the right order?

marenzo:
@starthal those are annotation of the previous version (0.6.x), and this PR is provided same serialization ability of array format also in the new client (>=0.8.5).

starthal:
If I use version 0.8.5 what do I need to do in order to (de)serialize a POJO in the array format? Do I need to be using Jackson?

marenzo:
@starthal yes, you should use msgpack-jackson package, and check the following doc:
https://github.com/msgpack/msgpack-java/tree/develop/msgpack-jackson

PR #359, opened 2016-04-08
Fix packString 

xerial:
While testing the config for disabling STR8 output, I found an unreachable code path in packString. This PR fixes the usage of CharsetEncoder used in packString.

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
@@ -471,6 +471,11 @@ public class MessagePacker
             // Underflow should be on to ensure all of the input string is encoded
             return -1;
         }
+        // NOTE: This flush method does nothing if we use UTR8 encoder, but other general encoders require this
+        cr = encoder.flush(bb);
+        if (!cr.isUnderflow()) {
+            return -1;
+        }
         return bb.position() - startPosition;
     }
 
--- a/msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
@@ -449,6 +449,7 @@ public class MessagePacker
         if (encoder == null) {
             this.encoder = MessagePack.UTF8.newEncoder();
         }
+        encoder.reset();
     }
 
     private int encodeStringToBufferAt(int pos, String s)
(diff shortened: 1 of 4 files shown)


xerial:
@frsyuki Could you check this PR? I found CharsetEncoder.underflow is reported even if the entire input string is properly consumed. For example, here is a code of StringCoding.encode: 
```
                ce.reset();
                ByteBuffer bb = ByteBuffer.wrap(ba);
                CharBuffer cb = CharBuffer.wrap(ca, off, len);
                try {
                    CoderResult cr = ce.encode(cb, bb, true);
                    if (!cr.isUnderflow())
                        cr.throwException();
                    cr = ce.flush(bb);
                    if (!cr.isUnderflow())
                        cr.throwException();
                } catch (CharacterCodingException x) {
                    // Substitution is always enabled,
                    // so this shouldn't happen
                    throw new Error(x);
                }
```
which is used in String.getBytes of Java.

msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
```diff
@@ -466,7 +467,13 @@ private int encodeStringToBufferAt(int pos, String s)
                 throw new MessageStringCodingException(e);
             }
         }
-        if (cr.isUnderflow() || cr.isOverflow()) {
+        if (!cr.isUnderflow() || cr.isOverflow()) {
+            // Underflow should be on to ensure all of the input string is encoded
+            return -1;
+        }
+        // NOTE: This flush method does nothing if we use UTR8 encoder, but other general encoders require this
```
frsyuki: what's UTR8 encoder?

frsyuki: oh, I think you mean UT"F"8

xerial: ah. I'll fix this.

msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
```diff
@@ -457,6 +457,7 @@ private int encodeStringToBufferAt(int pos, String s)
         ByteBuffer bb = buffer.sliceAsByteBuffer(pos, buffer.size() - pos);
         int startPosition = bb.position();
         CharBuffer in = CharBuffer.wrap(s);
+        encoder.reset();
```
frsyuki: I think it's slightly better to move this line to `prepareEncoder` method.

frsyuki:
ok, I see......that's very good catch.
I confirmed it by using jruby.
I commented some. others look good!

xerial:
Applied the comment. Thanks!

PR #362, opened 2016-04-19
Minor improvement of pull request 361

komamitsu:


msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackSerializerFactory.java
```diff
@@ -25,11 +25,19 @@
         extends BeanSerializerFactory
 {
     /**
+     * Constructor for creating instances without configuration.
+     */
+    public MessagePackSerializerFactory()
+    {
+        super(null);
```
xerial: How about using `this(null)`?

PR #363, opened 2016-04-19
Fix MessageUnapcker.prepareNumberBuffer

frsyuki:
prepareNumberBuffer was assuming that a buffer chunk includes at least
`readLength` bytes of data. Maximum size of `readLength` is 8.
Therefore, a problem could always happen if MessageBufferInput returns
buffer chunks smaller than 8 bytes.


xerial:
Nice catch. I added test cases to reproduced this issue.

PR #364, opened 2016-04-24
Support complex type key in MessagePackGenerator

komamitsu:
For https://github.com/msgpack/msgpack-java/issues/313#issuecomment-212530342


komamitsu:
Hmm. I noticed we need to pass the same AnnotationIntrospector and so on to a new ObjectMapper here if we should support the complete compatibility with v0.6...
https://github.com/msgpack/msgpack-java/pull/364/files#diff-a82a088fabde061b3431135d54a8ed2fR248

PR #367, opened 2016-04-27
Use REPLACE mode when converting java string into UTF8.

xerial:


frsyuki:
probably better to add a comment like this:

frsyuki:
```
Here sets REPLACE because following exception has happened before:
org.msgpack.core.MessageStringCodingException: java.nio.charset.MalformedInputException: Input length = 1
	at org.msgpack.core.MessagePacker.encodeStringToBufferAt(MessagePacker.java:467) ~[msgpack-core-0.8.6.jar:na]
	at org.msgpack.core.MessagePacker.packString(MessagePacker.java:535) ~[msgpack-core-0.8.6.jar:na]

This happened on JVM 7. But no ideas how to reproduce.
```

frsyuki:
Looks good :+1:

PR #373, opened 2016-07-05
Ignoring unknown properties on de-serializing process

marenzo:
Related to #358

Having two entities, similar in fields except one may contain extra one, ObjectMapper will fail to de-serialize as it expects the payload to have exact same number of fields (properties) on the value type object.

Consider such case:
```
Pojo1 {
 field1;
 field2;
}

Pojo2 {
 field1;
}
```

Previous client could handle a case when de-serializing the payload of Pojo2 into Pojo1.

This is totally solvable by adding `@JsonIgnoreProperties` annotation, but since we already have `JsonArrayFormat` as helper for backward-compatibility, thought it also should be part of this utility.


komamitsu:
👍

komamitsu:
@marenzo Thanks!

marenzo:
@komamitsu thanks for merging :) could you ping me when you'll release new version?

komamitsu:
@marenzo We released 0.8.8 a few weeks ago that includes this feature.

PR #374, opened 2016-07-06
Make MessageUnpacker.hasNext overridable

miniway:
In a case when a client knows the exact number of elements,
it can use an optimized overriden MessageUnpacker which always returns true


xerial:
OK. LGTM

PR #375, opened 2016-07-20
fix 'Unexpected UTF-8 encoder state' error at packString

frsyuki:
A bug is that 1<<32 is 1 in Java. It must be 1L<<32.
A problem happens packString is called with a String that contains a string (longer than 1<<8 characters) and (shorter than 1<<16 characters) and (longer than 1<<16 bytes in UTF-8).


xerial:
LGTM

muga:
thank you for fixing👍

PR #376, opened 2016-07-27
MessagePacker implements Flushable via flush()

LoneRifle:
This trivial change will add the Flushable interface to MessagePacker, and is already implemented through flush(). This will be useful to callers who expect a Flushable.


xerial:
Merged. Thanks.

PR #377, opened 2016-07-27
Add ByteBuffer input support again

miniway:


msgpack-core/src/main/java/org/msgpack/core/buffer/ByteBufferInput.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+public class ByteBufferInput
+        implements MessageBufferInput
+{
+    private ByteBuffer input;
+    private boolean isRead = false;
+
+    public ByteBufferInput(ByteBuffer input)
+    {
+        this.input = checkNotNull(input, "input ByteBuffer is null");
```
frsyuki: I think it is better to call `slice()` as following to avoid this unexpected scenario:

* An user creates a new ByteBuffer and fill some data there
* The user creates a new `ByteBufferInput` with the ByteBuffer
* the user changes position of the ByteBuffer! That might be by another thread.
* ByteBufferInput is affected the change of position, unexpectedly

```
        this.input = checkNotNull(input, "input ByteBuffer is null").slice();
```

msgpack-core/src/main/java/org/msgpack/core/buffer/ByteBufferInput.java
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * Reset buffer. This method doesn't close the old resource.
+     *
+     * @param input new buffer
+     * @return the old resource
+     */
+    public ByteBuffer reset(ByteBuffer input)
+    {
+        ByteBuffer old = this.input;
+        this.input = input;
```
frsyuki: same with above: 

```
        this.input = checkNotNull(input, "input ByteBuffer is null").slice();
```

msgpack-core/src/main/java/org/msgpack/core/buffer/ByteBufferInput.java
```diff
@@ -0,0 +1,69 @@
...
+    private ByteBuffer input;
+    private boolean isRead = false;
+
+    public ByteBufferInput(ByteBuffer input)
+    {
+        this.input = checkNotNull(input, "input ByteBuffer is null");
+    }
+
+    /**
+     * Reset buffer. This method doesn't close the old resource.
```
frsyuki: `This method doesn't close the old resource.` this text seems unnecessary

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
```diff
@@ -415,6 +494,11 @@ public ByteBuffer sliceAsByteBuffer()
         return sliceAsByteBuffer(0, size());
     }
 
+    public boolean hasArray()
```
frsyuki: 👍

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
```diff
@@ -225,13 +263,43 @@ public static void releaseBuffer(MessageBuffer buffer)
...
+    /**
+     * Create a MessageBuffer instance from a given ByteBuffer instance
+     *
+     * @param bb
+     */
+    MessageBuffer(ByteBuffer bb)
+    {
+        if (bb.isDirect()) {
+            if (isUniversalBuffer) {
+                throw new IllegalStateException("Cannot create MessageBuffer from DirectBuffer");
```
frsyuki: let's include "on this platform" in the message so that people don't think it's not MessagePack's fault...

```
+                throw new IllegalStateException("Cannot create MessageBuffer from DirectBuffer on this platform");
```

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
```diff
@@ -415,6 +494,11 @@ public ByteBuffer sliceAsByteBuffer()
         return sliceAsByteBuffer(0, size());
     }
 
+    public boolean hasArray()
+    {
+        return base instanceof byte[];
```
frsyuki: how about using `base != null` instead that seems faster?

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
```diff
@@ -202,11 +219,32 @@ private static MessageBuffer newMessageBuffer(byte[] arr, int off, int len)
...
+            // generated to resolve one of the method references when two or more classes overrides the method.
+            return (MessageBuffer) mbBBConstructor.newInstance(bb);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
     public static void releaseBuffer(MessageBuffer buffer)
     {
         if (isUniversalBuffer || buffer.base instanceof byte[]) {
```
frsyuki: @xerial I think that type of `MessageBuffer.buffer` field can be `byte[]` instead of `Object`. no? If yes, this `buffer.base instanceof byte[]` can be `buffer.base != null`.

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
```diff
@@ -202,11 +219,32 @@ private static MessageBuffer newMessageBuffer(byte[] arr, int off, int len)
...
             // We have nothing to do. Wait until the garbage-collector collects this array object
         }
+        else if (DirectBufferAccess.isDirectByteBufferInstance(buffer.base)) {
+            DirectBufferAccess.clean(buffer.base);
```
frsyuki: @miniway shouldn't this be `buffer.reference` because `base` is null?

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
xerial: @frsyuki If we never use this class for reading `float[]`, long[]`, etc. Using `byte[]` is ok.  If we need to read byte arrays more than 2GB, leaving this to Object is good since we can put long[] here so that we can read 8 (long byte size) * 2GB (2^31) = 16GB data.

frsyuki: I see...but I think it's too early to think about it. Exception handling becomes more complicated with consideration of long[] there.

frsyuki:
@miniway I wrote code for above comments at #378.
@xerial could you also review #378? Especially around improvements of exception handling and documents.

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
xerial: 2GB is not so huge, so we should think about it if we are going to support huge message pack based DataFrame in memory. 

The index of getXXX(index) is using int, so we already have a technical limit of the accessible buffer range, but this should not limit the underlying buffer size; if we use `long[]` we can create a MessageBuffer slice (2GB range) using on 16GB memory buffer.

xerial:
sure. will do

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
miniway: good catch! it should be the `reference`

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
miniway: I prefer `base != null` also.

miniway:
I've updated the PR based on the comments

PR #387, opened 2016-09-23
resetDecoder should be called once per decoding

miniway:
When multiple MessageBuffer involved at `unpackString` first bytes can be removed because `resetDecoder` is called multiple times.

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/buffer/SequenceMessageBufferInput.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/buffer/SequenceMessageBufferInput.java
@@ -1,0 +1,81 @@
+//
+// MessagePack for Java
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+//
+package org.msgpack.core.buffer;
+
+import java.io.IOException;
+import java.util.Enumeration;
+
+import static org.msgpack.core.Preconditions.checkNotNull;
+
+/**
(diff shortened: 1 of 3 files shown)


msgpack-core/src/main/java/org/msgpack/core/buffer/SequenceMessageBufferInput.java
```diff
@@ -0,0 +1,80 @@
...
+{
+    private Enumeration<? extends MessageBufferInput> sequence;
+    private MessageBufferInput input;
+
+    public SequenceMessageBufferInput(Enumeration<? extends MessageBufferInput> sequence)
+    {
+        this.sequence = checkNotNull(sequence, "input sequence is null");
+        try {
+            nextInput();
+        } catch (IOException ignore) {
```
xerial: code style check error: https://travis-ci.org/msgpack/msgpack-java/jobs/162190720#L507

msgpack-core/src/test/scala/org/msgpack/core/MessageUnpackerTest.scala
```diff
@@ -330,21 +360,28 @@ class MessageUnpackerTest extends MessagePackSpec {
       new SplitTest {val data = testData3(30)}.run
     }
 
-    "read numeric data at buffer boundary" taggedAs("boundary2") in {
+    "read data at buffer boundary" taggedAs("boundary2") in {
```
xerial: This test case is intended for checking a bug on unpackInteger that found in the past. We should have another test case for string.

xerial:
Merged in #388

msgpack-core/src/main/java/org/msgpack/core/buffer/SequenceMessageBufferInput.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * {@link MessageBufferInput} adapter for {@link MessageBufferInput} Enumeration
+ */
+public class SequenceMessageBufferInput
+        implements MessageBufferInput
+{
+    private Enumeration<? extends MessageBufferInput> sequence;
+    private MessageBufferInput input;
+
+    public SequenceMessageBufferInput(Enumeration<? extends MessageBufferInput> sequence)
```
frsyuki: can we use Iterator instead of Enumeration? or how about moving this to test package only for now?

xerial: Agreed. `Iterator<MessageBufferInput>` would work. And this should be in test package.

PR #389, opened 2016-09-23
Mimic SequenceInputStream interface more in SequenceMessageBufferInput

frsyuki:
I commented as https://github.com/msgpack/msgpack-java/pull/387#pullrequestreview-1377376 but I changed my mind: I think it's ok to keep SequenceMessageBufferInput as a mimic of SequenceInputStream. We can create another class if necessary.

A reason is that behavior of `SequenceMessageBufferInput.close` (which is same with`SequenceInputStream.close`) is surprising, in my opinion. I expect that it closes input sequence without reading buffers. But they consume all of them. However, change of the behavior is also surprising because its behavior becomes inconsistent from SequenceInputStream.

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/buffer/SequenceMessageBufferInput.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/buffer/SequenceMessageBufferInput.java
@@ -16,6 +16,8 @@
 package org.msgpack.core.buffer;
 
 import java.io.IOException;
+import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


frsyuki:
well, rather better idea is to move `SequenceMessageBufferInput` to test package. I think there're little demand for it. How do you think, @xerial?

xerial:
@frsyuki Just move it to test package. I think there will be no handy way to close a sequence of streams while managing there lifecycle.

Issue #398, opened 2016-10-22
Better Deserialization of Extension Types with Jackson


Chad-:
If you do a `readTree` that contains an extension type, the resulting tree will look something like (for example) `[1, 2, [MessagePackExtensionType, MessagePackExtensionType']`. In order to make this tree usable you need to traverse the whole tree converting any `MessagePackExtensionType` object to the real object type. It would be much better if reading the tree instead resulted in `[1, 2, [RealClass, RealClass]]`.

Ideally I would like to be able to do something along the lines of

```
factory = new MessagePackFactory
factory.register(typeId, class, encoder, decoder)
```

and have MessagePackParser check if a type is registered to return the correct object.

clarkdo:
+1

komamitsu:
@Chad- @clarkdo  Does this feature satisfy your request? https://github.com/msgpack/msgpack-java/pull/408/files#diff-da72e095f423bf63a2b05b8196b564dfR720

```
    ExtensionTypeCustomDeserializers extTypeCustomDesers = new ExtensionTypeCustomDeserializers();
    extTypeCustomDesers.addTargetClass((byte) 17, NestedListComplexPojo.class);
    extTypeCustomDesers.addTargetTypeReference((byte) 99, new TypeReference<Map<String, Integer>>() {});
    extTypeCustomDesers.addCustomDeser((byte) 31, new ExtensionTypeCustomDeserializers.Deser() {
                @Override
                public Object deserialize(byte[] data)
                        throws IOException
                {
                    if (Arrays.equals(data, new byte[] {(byte) 0xCA, (byte) 0xFE, (byte) 0xBA, (byte) 0xBE})) {
                        return "Java";
                    }
                    return "Not Java";
                }
            }
    );
    ObjectMapper objectMapper =
            new ObjectMapper(new MessagePackFactory().setExtTypeCustomDesers(extTypeCustomDesers));
```

komamitsu:
Merged the above feature. I hope it solves this request.

clarkdo:
@komamitsu thanks. It should be helpful.

PR #408, opened 2017-01-18
Jackson typebase ext deser

komamitsu:
For https://github.com/msgpack/msgpack-java/issues/398


komamitsu:
Hmm. This implementation works in `ObjectMapper#readValue` but doesn't work in `ObjectMapper#readTree` because `ObjectMapper#_readMapAndClose` runs with `ObjectMapper#JSON_NODE_TYPE` through `ObjectMapper#readTree`. In that case, `MessagePackExtensionType.TypeBasedDeserializer` isn't called at all...

PR #425, opened 2017-06-17
Add MessageBufferPacker#getSize()

isopov:
Without this method it seems that the only option to know size of data written to MessageBufferPacker prior to accessing it is to use toByteArray().length (too much copying) or toMessageBuffer().size() - still copying in case of much data.

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/MessageBufferPacker.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessageBufferPacker.java
@@ -123,4 +123,12 @@ public class MessageBufferPacker
         }
         return getArrayBufferOut().toBufferList();
     }
+
+    /**
+     * @return size of written data
+     */
+    public int getSize()
+    {
+        return getArrayBufferOut().getSize();
+    }
 }


komamitsu:
Why don't you use `MessagePacker#getTotalWrittenBytes`?

isopov:
That is easy - I reuse packer and clear it and getTotalWrittenBytes are not cleared.

```
		MessageBufferPacker packer = MessagePack.newDefaultBufferPacker();
		
		packer.packInt(Integer.MAX_VALUE);
		System.out.println(packer.getTotalWrittenBytes());
		
		packer.clear();
		System.out.println(packer.getTotalWrittenBytes());
```
Both times 5 is printed.

komamitsu:
I see. That makes sense.

BTW, the method name `getSize` sounds a bit too abstract to me since it's not clear that which size the method name means. How about `getOutputSize` or something?

And, it would be great if you add an unit test of the change (if possible).

xerial:
@isopov @komamitsu It sounds like MessageBufferPacker.clear should be fixed to reset the total written bytes value in MessagePacker. 

But adding getSize (I think getBufferSize() is a better name) seems OK because for MessageBufferPacker, the semantics of getBufferSize() is clear. Note that, however, MessagePacker also has its internal buffer, so there is no guarantee that the written byte size matches with getBufferSize.

xerial:
Merged in a51b520de77b109fb676904762def476cf03007f

PR #431, opened 2017-08-10
[WIP] Add add support for Timestamp type

frsyuki:
Implementation is still working in progress. 
Here is the spec of the new timestamp type: https://github.com/msgpack/msgpack/pull/209

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
@@ -32,6 +32,8 @@ import java.nio.charset.CharacterCodingException;
 import java.nio.charset.CharsetEncoder;
 import java.nio.charset.CoderResult;
 import java.nio.charset.CodingErrorAction;
+import java.time.Instant;
+import java.util.Date;
 
 import static org.msgpack.core.MessagePack.Code.ARRAY16;
 import static org.msgpack.core.MessagePack.Code.ARRAY32;
@@ -41,6 +43,7 @@ import static org.msgpack.core.MessagePack.Code.BIN8;
 import static org.msgpack.core.MessagePack.Code.EXT16;
 import static org.msgpack.core.MessagePack.Code.EXT32;
 import static org.msgpack.core.MessagePack.Code.EXT8;
+import static org.msgpack.core.MessagePack.Code.EXT_TIMESTAMP;
 import static org.msgpack.core.MessagePack.Code.FALSE;
 import static org.msgpack.core.MessagePack.Code.FIXARRAY_PREFIX;
 import static org.msgpack.core.MessagePack.Code.FIXEXT1;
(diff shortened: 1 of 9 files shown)


msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
```diff
@@ -32,6 +32,8 @@
 import java.nio.charset.CharsetEncoder;
 import java.nio.charset.CoderResult;
 import java.nio.charset.CodingErrorAction;
+import java.time.Instant;
```
xuwei-k: drop Java 7 support ?

https://github.com/msgpack/msgpack-java/blob/81d540d55d86862c775f16b9457c586c9127a790/.travis.yml#L12-L13

xerial: @frsyuki has confirmed it works in Java6 and Java7 even if Instant class is in the import statement. Only programs that uses Instance related methods will fail to compile.

xerial:
TODO:
- [ ] Add timestamp MessageFormat types?

mattbishop:
How is the timestamp feature coming along? I know the msgpack-js project is getting close to a release: https://github.com/kawanet/msgpack-lite/issues/76#issuecomment-347141538

xerial:
@mattbishop We haven't decided how to add above two things; whether to add MessageFormat 
 type and ValueType for describing timestamp.

In the msgpack spec, timestamp is just one of the Extention types. We think adding these two additional types will be convenient for timestamp users. However, adding these types is a big change to the API (since most of the users is using switch case statement for checking data type, so rewriting will be necessary by adding `case Timestamp: ...`), so we need to be carful so as not to break user programs that started to use timestamp types in future versions.

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
```diff
@@ -1232,6 +1246,70 @@ private String decodeStringFastPath(int length)
...
+                return (u32 & 0xffffffffL) * 1000L;
+            }
+            case 8: {
+                long data64 = readLong();
+                int nsec = (int) (data64 >>> 34);
+                long sec = data64 & 0x00000003ffffffffL;
+                return sec * 1000L + nsec / 1000L;
+            }
+            case 12: {
+                int nsec = readInt();
```
xerial: @frsyuki nsec is uint32, so &xffffffffL mask is necessary to convert negative int32 to uint32

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
```diff
@@ -1232,6 +1246,70 @@ private String decodeStringFastPath(int length)
...
+    public Instant unpackInstant()
+            throws IOException
+    {
+        ExtensionTypeHeader ext = unpackExtensionTypeHeader();
+        if (ext.getType() != EXT_TIMESTAMP) {
+            throw unexpectedExtension("Timestamp", EXT_TIMESTAMP, ext.getType());
+        }
+        switch (ext.getLength()) {
+            case 4: {
+                int u32 = readInt();
```
xerial: @frsyuki Need to convert to uint32 by applying 0xFFFFFFFFL mask

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
```diff
@@ -1232,6 +1246,70 @@ private String decodeStringFastPath(int length)
...
+                return Instant.ofEpochSecond(u32);
+            }
+            case 8: {
+                long data64 = readLong();
+                int nsec = (int) (data64 >>> 34);
+                long sec = data64 & 0x00000003ffffffffL;
+                return Instant.ofEpochSecond(sec, nsec);
+            }
+            case 12: {
+                int nsec = readInt();
```
xerial: This also needs to be converted to uint32 with 0xFFFFFFFFL mask

gustf:
Hi there, any progress on this feature?

xerial:
Continuing the work at #431

xerial:
Closing this PR in favor of #565

PR #435, opened 2017-10-10
expose method to create parser from direct memory

arnaudroger:
Useful to parse message coming in from netty for example without creating a lot of garbage.
Also expose the method to create a parser from a sub array.

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/buffer/ByteBufferInput.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/buffer/ByteBufferInput.java
@@ -64,4 +64,16 @@ public class ByteBufferInput
     {
         // Nothing to do
     }
+
+    /**
+     * Create a ByteBufferInput on off heap memory
+     * @param address the address
+     * @param offset the offset
+     * @param length the length
+     * @return a new ByteBufferInput on the specified address
+     */
+    public static ByteBuffer directBuffer(long address, int offset, int length)
+    {
+        return DirectBufferAccess.newByteBuffer(address, offset, length, null);
+    }
 }
--- a/msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackFactory.java
+++ b/msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackFactory.java
@@ -22,6 +22,7 @@ import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.io.IOContext;
 import org.msgpack.core.MessagePack;
+import org.msgpack.core.buffer.ByteBufferInput;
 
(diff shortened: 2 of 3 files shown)


xerial:
@komamitsu Could you review this change? This seems a good improvement.

msgpack-core/src/main/java/org/msgpack/core/buffer/ByteBufferInput.java
```diff
@@ -64,4 +64,16 @@ public void close()
     {
         // Nothing to do
     }
+
+    /**
+     * Create a ByteBufferInput on off heap memory
+     * @param address the address
+     * @param offset the offset
+     * @param length the length
+     * @return a new ByteBufferInput on the specified address
```
komamitsu: This static method returns ByteBuffer not ByteBufferInput

msgpack-core/src/main/java/org/msgpack/core/buffer/ByteBufferInput.java
```diff
@@ -64,4 +64,16 @@ public void close()
...
+     */
+    public static ByteBuffer directBuffer(long address, int offset, int length)
+    {
+        return DirectBufferAccess.newByteBuffer(address, offset, length, null);
```
komamitsu: It doesn't seem to me that this method is related to ByteBufferInput.

@xerial Do you think this is the right place to put the method in?

msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackFactory.java
```diff
@@ -99,6 +101,22 @@ public JsonParser createParser(byte[] data)
...
 
     @Override
+    public JsonParser createParser(byte[] data, int offset, int length)
+            throws IOException, JsonParseException
+    {
+        IOContext ioContext = _createContext(data, false);
+        return _createParser(data, offset, length, ioContext);
+    }
+
+    public JsonParser createParser(long memoryAddress, int offset, int length)
```
komamitsu: As you might notice since this method doesn't have `@Override` annotation, Jackson doesn't support this kind of API, I think.

How do you use this method via Jackson's API? It would be great, if you add some unit tests to make it clear 😺

arnaudroger: I use the MessagePackFactory directly, currently using reflection.

msgpack-core/src/main/java/org/msgpack/core/buffer/ByteBufferInput.java
xerial: We can make DirectBufferAccess a public class instead of adding this method.

xerial: Actually this location is a bit weird.

PR #436, opened 2017-10-11
Improve readPayload performance

miniway:
`readPayload(MessageBuffer)` is usually faster than `readPayload(ByteBuffer)` by using `unsafe.copyMemory`


xerial:
Merged. Thanks for a nice improvement!

PR #437, opened 2017-10-14
Add MessageUnpacker.tryUnpackNil() method

frsyuki:
When a code deals with an Optional<T> value, a common way is to
serialize a Nil value if the value is absent. To deserialize it, we
check whether the next value is nil or not first. If it is nil, skip the
byte. Otherwise, read a value. For this common use case, tryUnpackNil
simplifies the deserialization code. It does "check whether the next
value is nil or not first. If it is nil, skip the byte" in one method
call.


xerial:
LGTM

PR #439, opened 2017-10-22
fix java 9 build

xuwei-k:


xerial:
Merged. Thanks

xerial:
We may need to upgrade to sbt 1.0.4

PR #443, opened 2017-11-29
Upgrade to sbt-1.0.4 

xerial:
For better Java9 support. 
- Migrate to sbt 1.0.4, which has better Java9 build support
- Upgraded sbt-plugins for 1.0.4
- Introduced scalafmt for Scala code formatting
- Removed java7 test on Travis, but msgpack-java is still built for Java 7 (`-target 1.7`)
  - For Java7 testing, we should create a pre-built testing program and run it on JDK7


xerial:
Found one performance regression:
```
[info] - should be faster than msgpack-v6 read value *** FAILED ***
[info]   0.0027447835408163264 was not less than or equal to 0.0022912134897959185 (MessageUnpackerTest.scala:595)
```

xerial:
The performance regression never occurs in my local machine. It seems a problem of running test code inside a container of Travis. That test should be more lenient (e.g., just show warning)

xerial:
@komamitsu Could you check this PR? (cc: @xuwei-k)
This fixes recent build failures on Travis, and general updates of sbt plugins to the latest version. 
FYI. sbt-1.0.4 supports Scala 2.12 and Java9. So we no longer need to use Scala 2.10.

README.md
```diff
@@ -74,6 +74,7 @@ Here is a list of sbt commands for daily development:
 > findbugs                                 # Produce findbugs report in target/findbugs
 > jacoco:cover                             # Report the code coverage of tests to target/jacoco folder
 > jcheckStyle                              # Run check style
+> scalafmt                                 # Reformat Scala codes
```
xuwei-k: https://github.com/lucidsoftware/neo-sbt-scalafmt/blob/1.14/README.md#usage

> scalafmt       # format compile sources
> test:scalafmt  # format test sources
> sbt:scalafmt   # format .sbt source

The `scalafmt` command format only compile sources

.travis.yml
```diff
@@ -5,25 +5,27 @@ cache:
...
   include:
-  - dist: trusty
+  - env: PROJECT=checkstyle
+    jdk: oraclejdk8
+    script:
+    - ./sbt jcheckStyle
+  - env: PROJECT=java8
+    jdk: oraclejdk8
+    script:
+    - ./sbt test
```
xuwei-k: add `sbt ";scalafmt::test ;test:scalafmt::test ;sbt:scalafmt::test"` ?

https://github.com/lucidsoftware/neo-sbt-scalafmt/blob/1.14/README.md#usage
https://github.com/sbt/sbt/blob/v1.0.4/.travis.yml#L20


> To ensure everything is formatted, and fail if it is not (e.g. as a CI step),
>
> `>` scalafmt::test      # check compile sources
> `>` test:scalafmt::test # check test sources
> `>` sbt:scalafmt::test  # check .sbt sources

xuwei-k: Ah, Maybe I was wrong. You added `addSbtPlugin("com.geirsson" % "sbt-scalafmt"    % "1.3.0")`, not neo-sbt-scalafmt 🙇

README.md
xuwei-k: https://github.com/msgpack/msgpack-java/pull/443/files#r153711477

.travis.yml
xerial: Right. sbt-scalafmt reformats everything including main, test, *.sbt, etc.

xerial:
Thanks for checking! @komamitsu @xuwei-k

PR #444, opened 2017-11-29
Add 0.8.14 release notes

xerial:


xerial:
@frsyuki I'll release 0.8.14. Then we should start timestamp support in the next version.

frsyuki:
LGTM!!

PR #452, opened 2018-01-10
update sbt plugins

xuwei-k:
scalafmt plugin commands changed. see
- https://github.com/scalameta/scalafmt/pull/1085
- https://github.com/scalameta/scalafmt/blob/v1.4.0/scalafmt-sbt/src/main/scala/org/scalafmt/sbt/ScalafmtPlugin.scala#L18-L32


xerial:
OK. We can also use sbt-1.1.0

PR #456, opened 2018-02-07
remove "-target:jvm-1.7" from scalacOptions

xuwei-k:
Scala 2.12 does not support "-target:jvm-1.7"

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -15,7 +15,7 @@ val buildSettings = Seq[Setting[_]](
     Tags.limit(Tags.Test, 1)
   ),
   // JVM options for building
-  scalacOptions ++= Seq("-encoding", "UTF-8", "-deprecation", "-unchecked", "-target:jvm-1.7", "-feature"),
+  scalacOptions ++= Seq("-encoding", "UTF-8", "-deprecation", "-unchecked", "-feature"),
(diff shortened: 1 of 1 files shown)


xerial:
OK. Then we need to drop java7 support totally in upcoming version.

@komamitsu @frsyuki 
Should we bump the version number to 0.9.x before dropping Java7 support?

xuwei-k:
msgpack-java already use sbt 1.x (which does not support Java 7)

xuwei-k:
> but msgpack-core-0.8.14.jar still built for JVM7 (class version 0.51):

Yes, sbt 1.x can generate JVM7 byte code. but sbt 1.x and Scala 2.12 does not run on JVM 8.
This pull request changes `scalacOptions` not `javacOptions`. `scalacOptions` is different from `javacOptions`.
This change has no effect (just suppress warnings) because Scala 2.12 compiler says follows when set `-target:jvm-1.7`

```
[warn] -target is deprecated: -target:jvm-1.7 is deprecated, forcing use of jvm-1.8
```

xerial:
@xuwei-k Thanks for making it clear. I got confused scalacOptions and javacOptions.

xerial:
Merged.

Issue #458, opened 2018-02-12
MessageBuffer.sliceAsByteBuffer() method has old values after MessageBufferPacker.clear()


isopov:
Consider the following code:
```
MessageBufferPacker packer = MessagePack.newDefaultBufferPacker();
packer.packInt(1);
packer.flush();
packer.clear();

packer.packInt(2);

MessageBuffer buffer = packer.toBufferList().get(0);
System.out.println(Arrays.toString(packer.toByteArray()));
System.out.println(Arrays.toString(buffer.toByteArray()));
byte[] array = Arrays.copyOf(buffer.sliceAsByteBuffer().array(), buffer.size());
System.out.println(Arrays.toString(array));
```

I expected it to print the same value 3 times, but its output is:

```
[2]
[2]
[1]
```

AFAIU there is a bug somewhere - in clear() implementation or somewhere else. I think it is in clear() because removing `flush()` produces the following output:

```
[1, 2]
[1, 2]
[1, 2]
```

It seems that `clear()` has no effect at all in this case.

xerial:
@isopov Current MessagePacker interface is not designed for allow rewriting, because the buffer)(internally multiple buffers are used) might be already flushed out to the output target.

If you need to reuse buffer, MessagePacker.reset is the method for that:
https://static.javadoc.io/org.msgpack/msgpack-core/0.8.14/org/msgpack/core/MessagePacker.html#reset-org.msgpack.core.buffer.MessageBufferOutput-

For BufferPacker, fixes in #459 looks OK to have. I'll merge it. Thanks

PR #459, opened 2018-02-12
add MessagePacker.clear() method to clear position

isopov:
Fix for #458


xerial:
Thanks for the fix! Merged.

PR #460, opened 2018-02-13
replace ArrayList and defensive coping with single unmodifiableList

isopov:


isopov:
Build failed with this check:

```
-unpack performance	total:14.217 sec., count:  100, avg:0.142 sec. , core avg:0.140 sec. , min:0.084 sec. , max:0.415 sec. 
  -v6             	total:0.300 sec. , count:  100, avg:3.003 msec., core avg:2.981 msec., min:1.824 msec., max:6.346 msec.
  -v7-array       	total:0.324 sec. , count:  100, avg:3.238 msec., core avg:3.215 msec., min:1.876 msec., max:6.911 msec.
  -v7-array-buffer	total:0.320 sec. , count:  100, avg:3.203 msec., core avg:3.178 msec., min:1.871 msec., max:6.985 msec.
  -v7-direct-buffer	total:0.321 sec. , count:  100, avg:3.211 msec., core avg:3.187 msec., min:1.902 msec., max:6.954 msec.
[MessageUnpackerTest] v7-array 0.003214773591836735 is slower than v6 0.002980928377551021
[MessageUnpackerTest] v7-array-buffer 0.0031776363673469386 is slower than v6 0.002980928377551021
[info] - should be faster than msgpack-v6 read value *** FAILED ***
[info]   0.0031865427448979594 was not less than or equal to 0.002980928377551021 (MessageUnpackerTest.scala:602)
```

Seems to be not connected with my changes.

xerial:
@isopov What is the purpose of this change? I don't think creating a view of the internal ArrayList is necessary here.

isopov:
@xerial I believe unmodifiableList
- is lighter than ArrayList
- makes it more clear that this method is for reading bytes without extra copying and result list is not intended for modification (is it?)
- can be reused (maybe it has more sense to lazy create it, or to not reuse at all)

xerial:
@isopov OK.
Returning view is not ideal since the list content can be changed. So even if it has a slight overhead (I think creating an array list of buffer references is a cheap operation), I'd like to create a snapshot of the internal buffers. In terms overhead, crating a new listView object for each ArrayBufferOutput instantiation adds its own (really small) overhead, while toBufferList will not used for general applications.  

For example, toBufferList is used like this:
https://github.com/embulk/embulk-executor-mapreduce/blob/a41d5370d0d6b8f0855bd660cb183f8f52037976/embulk-executor-mapreduce/src/main/java/org/embulk/executor/mapreduce/PageWritable.java#L56

If we return a reference (not a snapshot), the application code like this must consider the change of the contents. I think it's annoying for users. So creating a snapshot (shallow copy) is better. 

For your second point, we'd like to return an ImmutableList (available in Guava), but we also don't want to add Guava into the dependency.

isopov:
@xerial Since returning ArrayList is a thoughtful choice and not a glitch (I thought so) it is perfectly OK to decline this PR, I think.

BTW - I do not see how the code you referenced would broke with this change...

xerial:
@isopov Thanks for understanding. The above code doesn't run any code that can modify the buffer list (e.g., packing extra values), so practically there is no problem. But the first part is counting the buffer size total and in the second part it's dumping the contents, so we need a guarantee the contents are the same (i.e., immutable).

Let me close this PR. Thanks.

PR #477, opened 2018-05-25
use MessageBuffer constructor reflection only when required

joelnn:
Invoking the MessageBuffer constructor showed up as a hotspot in my profiling. There is no need for it when we are only using MessageBuffer itself, not one of the subclasses.

This patch improves unpackValue performance for my data by about 25%.


xerial:
@joelnn We prefer not mixing two or more changes (Unpacker and reflection). Could you make different PRs so that we can check these changes individually? 

Avoiding reflection is OK, but I'm not sure about MessageUnpacker changes, so I cannot merge this PR.

PR #478, opened 2018-06-09
use MessageBuffer constructor reflection only when required

joelnn:
Invoking the MessageBuffer constructor showed up as a hotspot in my profiling. There is no need for it when we are only using MessageBuffer itself, not one of the subclasses.

This patch improves unpackValue performance for my data by about 25%.

(previously #477)


joelnn:
Thoughts on merging this?

PR #479, opened 2018-06-09
avoid creating MessageBuffer needlessly, convert degenerate switch to if

joelnn:
Avoid creating MessageBuffer just as a wrapper, MessageBuffer.getBytes() is there for that.

(previously #477)


xerial:
OK. Looks good. Thanks!

Issue #508, opened 2019-08-09
Nested serialization throws NullPointerException


hkalina:
When serializing object whose getter use MessagePack serialization, the inner serialization breaks the outer serialization and it throws NullPointerException:
```
java.lang.NullPointerException
	at com.fasterxml.jackson.core.util.ByteArrayBuilder.write(ByteArrayBuilder.java:224)
	at org.msgpack.core.buffer.OutputStreamBufferOutput.write(OutputStreamBufferOutput.java:78)
	at org.msgpack.core.buffer.OutputStreamBufferOutput.writeBuffer(OutputStreamBufferOutput.java:71)
	at org.msgpack.core.MessagePacker.flushBuffer(MessagePacker.java:306)
	at org.msgpack.core.MessagePacker.flush(MessagePacker.java:281)
	at org.msgpack.core.MessagePacker.close(MessagePacker.java:296)
	at org.msgpack.jackson.dataformat.MessagePackGenerator.close(MessagePackGenerator.java:509)
	at com.fasterxml.jackson.databind.ObjectMapper._configAndWriteValue(ObjectMapper.java:3911)
	at com.fasterxml.jackson.databind.ObjectMapper.writeValueAsBytes(ObjectMapper.java:3244)
	at org.msgpack.jackson.dataformat.MessagePackFactoryTest.testNestedSerialization(MessagePackFactoryTest.java:62)
```
It tries to write into ByteArrayBuilder which is already released/closed.

Reproducible using following test: (feel free to include this or derivated into msgpack tests)
```java
    @Test
    public void testNestedSerialization() throws Exception
    {
        ObjectMapper objectMapper = new ObjectMapper(new MessagePackFactory());
        objectMapper.writeValueAsBytes(new OuterClass());
    }

    public class OuterClass
    {
        public String getInner() throws JsonProcessingException
        {
            ObjectMapper m = new ObjectMapper(new MessagePackFactory());
            m.writeValueAsBytes(new InnerClass());
            return "EFG";
        }
    }

    public class InnerClass
    {
        public String getName()
        {
            return "ABC";
        }
    }
```
Identical test using standard (JSON) ObjectMapper works ok - only MessagePack is affected.

zde:
FYI this is probably caused by storing the byte output buffer in a ThreadLocal in MessagePackGenerator. Should be either converted to a field, or properly saved/restored before use.

komamitsu:
Wow, thanks for reporting that. I'll check it around mid-next week

[...]

PR #509, opened 2019-08-14
Fix issue:508

komamitsu:
Fixes https://github.com/msgpack/msgpack-java/issues/508

This change makes MessagePackGenerator manage status of OutputStreamBufferOutput in ThreadLocal storage to avoid manipulating a shared OutputStreamBufferOutput by multiple MessagePackGenerator(s) in the same thread. With this change, lately generated MessagePackGenerator can detect the OutputStreamBufferOutput in ThreadLocal is already `in use` and  create a new non-shared OutputStreamBufferOutput.

But it can lead an issue that if a MessagePackGenerator leaves a OutputStreamBufferOutput open in ThreadLocal, succeeding MessagePackGenerator(s) can't use the OutputStreamBufferOutput in ThreadLocal since it's still `in use` and it would affect the performance and also it's kinda leak.

So in this change, I additionally introduced CleanerService that's a sort of garbage collection. It changes the status of OutputStreamBufferOutput to `not in use` when the MessagePackGenerator gets not used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: scala
 
+# With xenial, `Installing oraclejdk8` fails due to "Expected feature release number in range of 9 to 14, but got: 8"
+dist: trusty
+
 cache:
   directories:
     - $HOME/.m2/repository/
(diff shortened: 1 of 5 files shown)


komamitsu:
Hmm... I tested this change with high frequent instantiations of MessagePackGenerator for several minutes and I noticed it caused Full GC compared... I guess registering an object to ReferenceQueue causes this issue.

PR #510, opened 2019-08-25
Better class names instead of hardcoded strings

k0l0r3k99:
Using xyz.class.getName() is better than using "xyz" due to proguard and other obfuscators


xerial:
@k0l0r3k99 Using hard-coded strings is quite intentional to avoid generating TypeProfiles inside JVM. If we reference classes (e.g., MessageBufferU), the performance of MessageBuffer will be slower by 10% or more. 

Let me close this PR as we can't merge this for the performance reason. 

Anyway, thanks for the proposal to fix.

Issue #482, opened 2018-09-04
MessageBuffer unable to parse java.specification.version for java 10


arnaudroger:
java.specification.version for java 10 is "10" 
MessageBuffer is expecting a '.' and will then not use unsafe even if available
I suspect the same is true for 9 and 11.
a workaround is to override the property on the command line 
```
-Djava.specification.version=1.10
```

ppkarwasz:
It's fixed in 0.8.19.

PR #514, opened 2019-09-13
Improves MessageBuffer support for Java 11

ppkarwasz:
Switches the MessageBuffer implementation to MessageBufferU for Java versions at least 9,  which solves a couple of problems:

1. On Java 11 (OpenJDK) jdk.internal.ref.Cleaner#clean is not open to reflection and the static initializer of DirectBufferAccess throws an IllegalAccessException.
2. Access to java.nio.DirectByteBuffer may suffer the same problem in the future. For now we just get a warning.

Since MessageBufferU overrides almost all methods of MessageBuffer, while direct ByteBuffers register a cleaner by themself, I don't see why MessageBufferU(ByteBuffer) shouldn't accept direct buffers.

This should also solve bug number #482.

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
@@ -165,7 +165,8 @@ public class MessageBuffer
         }
     }
 
-    static int getJavaVersion() {
+    static int getJavaVersion()
+    {
         String javaVersion = System.getProperty("java.specification.version", "");
         int dotPos = javaVersion.indexOf('.');
         if (dotPos != -1) {
--- a/msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
@@ -47,6 +47,7 @@ public class MessageBuffer
 {
     static final boolean isUniversalBuffer;
     static final Unsafe unsafe;
+    static final int javaVersion = getJavaVersion();
 
     /**
      * Reference to MessageBuffer Constructors
@@ -69,9 +70,6 @@ public class MessageBuffer
         int arrayByteBaseOffset = 16;
 
         try {
-            // Check java version
-            int javaVersion = getJavaVersion();
-
             boolean hasUnsafe = false;
             try {
                 hasUnsafe = Class.forName("sun.misc.Unsafe") != null;
@@ -163,7 +161,7 @@ public class MessageBuffer
         }
     }
 
-    static int getJavaVersion()
+    private static int getJavaVersion()
     {
         String javaVersion = System.getProperty("java.specification.version", "");
         int dotPos = javaVersion.indexOf('.');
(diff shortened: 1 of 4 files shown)


msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
```diff
@@ -97,12 +85,14 @@
             // Is Google App Engine?
             boolean isGAE = System.getProperty("com.google.appengine.runtime.version") != null;
 
-            // For Java6, android and JVM that has no Unsafe class, use Universal MessageBuffer
+            // For Java6, android and JVM that has no Unsafe class, use Universal MessageBuffer (based on ByteBuffer).
+            // Java9 onward doesn't allow to access Cleaner by reflection, so we use Universal MessageBuffer again.
```
xerial: The performance is the most important factor of msgpack-java. Insteaed of using MessageBufferU for Java9 or Java11, we should have a workaround for the Cleaner issue.

xerial:
@ppkarwasz Thanks for the suggestions. Unfortunately, however, using MessageBufferU for Java 9 or later cannot be our option as it will have a significant impact on the performance. We need to find a way to make DirectBufferAccess work on jdk9 or later.

msgpack-core/src/main/java/org/msgpack/core/buffer/MessageBuffer.java
ppkarwasz: I disabled the automatic switch to `MessageBufferU` on Java 9+.

ppkarwasz:
I removed the automatic switch to `MessageBufferU` on Java 9+ and disabled Travis tests without the `msgpack.universal-buffer` property. Those tests where using `MessageBufferU` anyway, since the original code gave `isJavaAtLeast7 == false` on Java 11.

With these last changes this PR may still serve a purpose: the original code throws an exception, when we instantiate a `MessageBufferU` using a direct `ByteBuffer`. This behavior is, IMHO, too cautious. At least on Java 8+ direct `ByteBuffer`s can clean up themselves so using direct buffers with `MessageBufferU` would not cause memory leaks.

In our use case the support for direct buffers is essential: we feed `MessageUnpacker` using **netty** (which almost always gives us direct buffers).

ppkarwasz:
> We need to find a way to make DirectBufferAccess work on jdk9 or later.

I might adapt **netty-common**'s solution to the `Cleaner` problem and provide a PR.

xerial:
@ppkarwasz Yes. We also use netty or jetty-provided ByteBuffer with msgpack. What is netty-common's approach for supporting jdk9? If it's reasonable, I'd like to use it in this PR instead of ignoring jdk11 test for MessageBuffer.

ppkarwasz:
In Java 9+ `sun.misc.Unsafe` gets an `invokeCleaner` method which is accessible through reflection.

ppkarwasz:
Done. Now it should work on both Java 8 and Java 11.

Summarizing, this PR:
* allows direct buffers, when "msgpack.use-universal=true",
* adds direct buffer access on Java 9+ runtimes.

xerial:
@ppkarwasz Thanks for your work. We will take over the remaining task (if necessary) as we also need to support Java11. (cc: @miniway)

miniway:
It looks very good to me. Also we could encapsulate the Cleaner in a separate class in the future

xerial:
It also looks good to me. Let me merge it and release a new version.

PR #520, opened 2020-05-10
write an ArrayBufferOutput to a WritableByteChannel

richardstartin:
Makes it possible to write the contents of an `ArrayBufferOutput` to a `WritableByteChannel` without materialising an `ArrayList<MessageBuffer>`


richardstartin:
ping @xerial

PR #524, opened 2020-06-16
Add GitHub Action workflow (x64) and Travis (arm64) build 

xerial:
- Use GitHub Actions for regular CIs
- Use Travis CI for testing on Arm64 with openjdk11 https://docs.travis-ci.com/user/multi-cpu-architectures/ It's an emulated environment running on x64 machine.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -32,7 +32,7 @@ jobs:
       - uses: olafurpg/setup-scala@v7
         with:
           java-version: adopt@1.11
-      - uses: actions/cache@v1
+      - uses: actions/cache@v2
         with:
           path: ~/.cache
           key: ${{ runner.os }}-jdk11-${{ hashFiles('**/*.sbt') }}
@@ -49,7 +49,7 @@ jobs:
       - uses: olafurpg/setup-scala@v7
         with:
           java-version: adopt@1.8
-      - uses: actions/cache@v1
+      - uses: actions/cache@v2
         with:
           path: ~/.cache
           key: ${{ runner.os }}-jdk8-${{ hashFiles('**/*.sbt') }}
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -0,0 +1,60 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: jcheckstyle
+        run: ./sbt jcheckStyle
+  test_jdk11:
+    name: Test JDK11
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: olafurpg/setup-scala@v7
```
yuokada, line 32: If the AdoptJDK installation is supported in setup-java action, it's better to replace with it.
https://github.com/actions/setup-java

.github/workflows/CI.yml
```diff
@@ -0,0 +1,60 @@
...
+        with:
+          java-version: adopt@1.11
+      - uses: actions/cache@v2
+        with:
+          path: ~/.cache
+          key: ${{ runner.os }}-jdk11-${{ hashFiles('**/*.sbt') }}
+          restore-keys: ${{ runner.os }}-jdk11-
```
yuokada, line 39: Is this key correct?

.github/workflows/CI.yml
xerial: msgpack-java's test is written in Scala, and also setup-scala can install natively-built sbt (csbt) using GraalVM. So let me keep using setup-scala.

.github/workflows/CI.yml
xerial: Yes. It has fallback mechanism https://circleci.com/docs/2.0/caching/#example-caching-configuration

yuokada: I read below document after I wrote above comment. And though I understood the specification of `restore-keys`, I forgot removing it.
https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows

PR #525, opened 2020-06-23
fix indexing bug in ValueFactory

Minh-Ng:
fix ValueFactory.newMap(Entry<Value, Value>...pairs), which is currently iterating by 2 when it should be iterating through each passed in pair.

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/value/ValueFactory.java
+++ b/msgpack-core/src/main/java/org/msgpack/value/ValueFactory.java
@@ -232,7 +232,7 @@ public final class ValueFactory
     public static MapValue newMap(Map.Entry<? extends Value, ? extends Value>... pairs)
     {
         Value[] kvs = new Value[pairs.length * 2];
-        for (int i = 0; i < pairs.length; i += 2) {
+        for (int i = 0; i < pairs.length; ++i) {
             kvs[i * 2] = pairs[i].getKey();
             kvs[i * 2 + 1] = pairs[i].getValue();
         }


xerial:
I believe the original code is already correct because `[i * 2]` index is already used. Could you add a test case so that we can confirm your fix is valid?

Minh-Ng:
Test code on the project itself looks pretty light. In any case we can take a visual test case of 
Map.Entry<? extends Value, ? extends Value> pairs = [Map.Entry<Value, Value> a, Map.Entry<Value, Value> b]. 

We want to process each pair, but the index you are referring to is filling the new array. So the for loop is actually trying to read non-existent indices of the pairs input array.

Minh-Ng:
In other words try and replace the for loop condition with

```
int fillIndex = 0;
int[] newArray = new int[pairs.length * 2];
for (Map.Entry<? extends Value, ? extends Value> entry : pairs) {
     newArray[fillIndex] = pair.getKey();
     newArray[fillIndex + 1] = pair.getValue();
     fillIndex += 2;
}
```

xerial:
ah. Ok. You are right. I was looking at the wrong code. We need to iterate the original input one by one.

If you can add a very simple test for this, it will be helpful.

Minh-Ng:
Is it just me or there are no unit tests for src/test/java? I just came here from a nullpointerexception  from an external library using messagepack core as a dependency. I'm not so familiar with this source.

Minh-Ng:
```
Map.Entry<Value, Value> entry1 = new AbstractMap.SimpleImmutableEntry<>(ValueFactory.newInteger(10L), ValueFactory.newInteger(100L));
Map.Entry<Value, Value> entry2 = new AbstractMap.SimpleImmutableEntry<>(ValueFactory.newInteger(100L), ValueFactory.newInteger(10L));

ImmutableMapValue newMap = ValueFactory.newMap(entry1, entry2);

String shouldNotThrowNPE = newMap.toString();
assertTrue(shouldNotThrowNPE != null && !"".equals(shouldNotThrowNPE));
assertEquals(2, newMap.asMapValue().entrySet().size())
```

xerial:
Although some test is lacking, we should merge this fix. Thanks.

PR #528, opened 2020-08-30
use newer version of jackson (2.10.5)

pjfanning:
https://github.com/msgpack/msgpack-java/issues/529 -- jackson 2.11.2 causes some test failures

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -116,7 +116,7 @@ lazy val msgpackJackson =
         "org.msgpack.jackson.dataformat"
       ),
       libraryDependencies ++= Seq(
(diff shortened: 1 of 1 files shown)


xerial:
@komamitsu We've confirmed upto jackson 2.10.x, it's ok to use with TD's internal ecosystem. Jackson 2.11.x is still not tested (e.g., in Presto, etc.)

komamitsu:
@pjfanning Thanks for the PR!

komamitsu:
@xerial I see. Let me know when you test 2.11 with Presto and so on.

PR #538, opened 2020-12-13
Fix GitHub Action CI

komamitsu:


xerial:
Thanks for the fix. jdk8 test also needs to use the latest version of setup-scala :)

PR #545, opened 2021-02-14
Fixes #544: Fix a bug in reading EXT32 with 2GB size

### LINKED ISSUE #544: Add boundary check to skipPayload(xxx)
We need to handle when an invalid value is passed to skipPayload:
https://github.com/msgpack/msgpack-java/blob/develop/msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java#L1476-L1489

xerial:


msgpack-core/src/test/scala/org/msgpack/core/InvalidDataReadTest.scala
```diff
@@ -0,0 +1,20 @@
...
+/**
+ *
+ */
+class InvalidDataReadTest extends MessagePackSpec {
+
+  "Reading long EXT32" in {
+      val msgpack = createMessagePackData(p => p.packExtensionTypeHeader(MessagePack.Code.EXT32, Int.MaxValue))
+      val u = MessagePack.newDefaultUnpacker(msgpack)
+      try {
+        intercept[MessageInsufficientBufferException] {
```
komamitsu, line 12: The expectation of this test case is unclear to me. I guess you want to avoid 2GB memory allocation in this test, though. How about adding a comment to describe your intention?

xerial: Added more comment to explain why we need to use this a bit weird test code

komamitsu (approved):
👍

Issue #542, opened 2020-12-24
MessageBuffer allows unsafe out-of-bounds reads and writes


panicbit:
The `get` and `put` family of functions on `MessageBuffer` are all missing bounds checks and thus can read and write out of bounds.

E.g. in the following example, the read `value` is either an undefined value or triggers a segfault:
```kotlin
val unpacker = MessagePack.newDefaultUnpacker(ByteArray(0))
val buffer = unpacker.readPayloadAsReference(0)
val value = buffer.getInt(9000000)

println("Value: $value")
```

Since this unsafety is exposed publicly and not documented, it can be quite dangerous.
If the ability to skip the bounds check is an intended feature, then I'd suggest to name the methods accordingly (e.g. giving them an "unsafe" prefix). Having a set of functions that does bounds checking by default probably does not hurt either.

komamitsu:
@xerial Can you take a look at this?

xerial:
@panicbit Thanks for the report. Although this method has been exposed, we have no intention to let normal users who don't have the internal implementation knowledge of msgpack-java to rely on this functionality. As this method is already used in some of our dependencies, renaming this method to readUnsafeXXX cannot be our option. And also, adding boundary checks to MessageBuffer methods might also have non-negligible performance overhead, so let us just fix the documentation as in #546

PR #546, opened 2021-02-14
#542: Add a warning note for the usage of MessageUnpacker.readPayloadAsReference

xerial:


komamitsu (approved):
The user of this method can know the buffer length encoded in the header. Even with a wrong length in header encoded by an attacker, it should be safe as long as the user of this method implements safely.

PR #548, opened 2021-05-11
Prevent having two different unpacking methods

cyberdelia:
Prevent having two different unpacking methods in the unpacker: `unpackValue` exclusively use arrays for both arrays and maps , and `unpackValue(Variable)`  uses list and map. This will prevent having two different behaviour when unpacking values and prevent some code duplication.


cyberdelia:
Well, I'm still worried by the lack of tests failing because of my improper logic copy/pasta.

PR #552, opened 2021-05-11
Update build settings to use latest version of sbt and plugins

xerial:
- Updated sbt and its plugins
- Fixed build.sbt to use slash syntaxes https://www.scala-sbt.org/1.x/docs/Migrating-from-sbt-013x.html#slash
- Upgraded sbt-sonatype to the latest version to have more release stability.
- Code coverage test doesn't work well with Jacoco. We need another coverage solution


xerial:
FYI: @komamitsu

PR #565, opened 2021-05-11
Timestamp support

xerial:
Follow-up work of #431

This PR is based on the commits at #431 and the timestamp implementation of airframe-msgpack https://github.com/wvlet/airframe/tree/master/airframe-msgpack/src/main/scala/wvlet/airframe/msgpack/spi

- [x] Add pack/unpackTimestamp. (Instant, milliseconds, sec+nano interface). Don't need to add other interfaces like unpackTimestamp(java.util.Date), unpackTimestamp(DateTime), etc. as these datetime instances can be converted into milliseconds or Instant at ease. Instant class requires JDK8 or later. So JDK7 users can't use timestamp but it should be ok as JDK7 is no longer maintained.
- [x] Add corner-case tests for various Timestamp values
- [x] Fix uint32 handling around nano seconds
- [x] Add Value support for Timestamp. Decided not to support Timestamp as the first-class ValueType because Timestamp requires reading 1 to 3 bytes for deciding ValueType while the other types only need the first 1 byte. 
- [x] Extend Variable to support reading Timestamp values
- [x] Enhance Variable tests 
- [x] Fix property-based testing. Needed to migrate to AirSpec https://wvlet.org/airframe/docs/airspec. ScalaTests with ScalaCheck was not properly running forAll methods
- [x] Remove ScalaTest, xerial-java dependencies


msgpack-core/src/test/scala/org/msgpack/core/MessagePackTest.scala
```diff
@@ -50,549 +53,585 @@ class MessagePackTest extends MessagePackSpec {
...
 
+    // Corner-cases around uint32 boundaries
+    for (v <- Seq(
+           Instant.ofEpochSecond(Instant.now().getEpochSecond, 123456789L), // uint32 nanoseq (out of int32 range)
+           Instant.ofEpochSecond(-1302749144L, 0), // 1928-09-19T21:14:16Z
+           Instant.ofEpochSecond(-747359729L, 0), // 1946-04-27T00:04:31Z
+           Instant.ofEpochSecond(4257387427L, 0) // 2104-11-29T07:37:07Z
+         )) {
+      check(v, _.packTimestamp(v), _.unpackTimestamp())
+    }
```
xerial, line 616: These test cases fixes uint32 conversion bugs in #431

xerial:
@komamitsu I'll merge this and prepare 0.9.0 release.

dmikurube:
@xerial @komamitsu If `msgpack-java` goes to 0.9.0, not next 0.8, can we consider this type of change? 0.8 -> 0.9 is a good chance to include some incompatibility...

https://github.com/msgpack/msgpack-java/pull/571

PR #597, opened 2021-09-10
Specify the bufferSize of the ArrayBufferOutput

szhnet:
Use the bufferSize of PackerConfig to specify the bufferSize of the ArrayBufferOutput in MessageBufferPacker.


xerial:
Nice catch. Thanks

PR #621, opened 2022-01-18
Keep consistent read size after closing MessageUnpacker

okumin:
I found `MessageUnpacker#getTotalReadBytes` doesn't return correct size once I invoke `MessageUnpacker#close`.

That's because `MessageUnpacker#close` drops the current buffer without updating the read size
https://github.com/msgpack/msgpack-java/blob/v0.9.0/msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java#L1738-L1740

even though `MessageUnpacker#getTotalReadBytes` requires the size of the latest buffer.
https://github.com/msgpack/msgpack-java/blob/v0.9.0/msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java#L261

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
@@ -1735,6 +1735,7 @@ public class MessageUnpacker
     public void close()
             throws IOException
     {
+        totalReadBytes += position;
         buffer = EMPTY_BUFFER;
         position = 0;
         in.close();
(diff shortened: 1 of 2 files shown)


komamitsu:
@xerial ^ What do you think?

yuokada:
@komamitsu It's good to ping him via slack or twitter. IIRC, he doesn't watch notifications from github.

xerial:
Thanks for the fix. Yeah. `position` value needs to be reflected to the total read size before resetting to 0

PR #622, opened 2022-01-23
Fixed examples relative link in README

zbuster05:
On the [message pack](https://msgpack.org/index.html) mirror of the readme, since this is relatively linked, it attempts to route to the location https://msgpack.org/msgpack/msgpack-java/blob/develop/msgpack-core/src/test/java/org/msgpack/core/example/MessagePackExample.java which subsequently returns a 404. Statically linking the domain should fix this issue and look a lot more professional. I personally almost didn't use this implementation since I did not realize this was an unintentional side effect.

And yes, I'm making a PR to change one line in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,9 @@ dependencies {
 ### Integration with Jackson ObjectMapper (jackson-databind)
 
 msgpack-java supports serialization and deserialization of Java objects through [jackson-databind](https://github.com/FasterXML/jackson-databind).
-For details, see [msgpack-jackson/README.md](msgpack-jackson/README.md). The template-based serialization mechanism used in v06 is deprecated.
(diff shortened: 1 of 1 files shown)


oza:
@zbuster05 Thanks a lot for sending the PR. Good catch. I also found the same problem in the following lines. Could you fix them too?
- https://github.com/msgpack/msgpack-java/blob/develop/README.md?plain=1#L48
- https://github.com/msgpack/msgpack-java/blob/develop/README.md?plain=1#L50

zbuster05:
> @zbuster05 Thanks a lot for sending the PR. Good catch. I also found the same problem in the following lines. Could you fix them too?
> 
> * https://github.com/msgpack/msgpack-java/blob/develop/README.md?plain=1#L48
> * https://github.com/msgpack/msgpack-java/blob/develop/README.md?plain=1#L50

Fixed! Let me know if there are any other changes you'd like me to make.

xerial:
Thanks for the fix!

PR #658, opened 2022-06-22
Fixed UnpackerConfig.bufferSize to work

### LINKED ISSUE #657: Consuming a large amount of memory when unpacking huge Array and Map unexpectedly
There is a problem of unexpectedly consuming a large amount of memory when unpacking huge Array and Map.
For example, this code may causes an OutOfMemoryError.

```java
    public static void unpackHugeArray()
            throws IOException
    {
        MessageBufferPacker packer = MessagePack.newDefaultBufferPacker();
        packer.packArrayHeader(2147483647);
        packer.close();

        byte[] packedData = packer.toByteArray();
        MessageUnpacker unpacker = MessagePack.newDefaultUnpacker(packedData);
        unpacker.unpackValue();  // Caused by: java.lang.OutOfMemoryError
        unpacker.close();
    }
```

napo0703:
This will Resolve #657

It limits internal buffer size when unpacking Array and Map.
However, I think the default bufferSize of 8192 is too small.
I changed the default buffer size to 100MiB following [msgpack-python](https://github.com/msgpack/msgpack-python#security).

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
@@ -518,7 +518,7 @@ public class MessagePack
 
         private int stringSizeLimit = Integer.MAX_VALUE;
 
-        private int bufferSize = 8192;
+        private int bufferSize = 100 * 1024 * 1024;
 
         private int stringDecoderBufferSize = 8192;
 
@@ -744,8 +744,8 @@ public class MessagePack
         }
 
         /**
-         * When a packer is created with newUnpacker(OutputStream) or newUnpacker(WritableByteChannel), the stream will be
-         * buffered with this size of buffer (default: 8192).
+         * When a packer is created with {@link #newUnpacker(InputStream)} or {@link #newUnpacker(ReadableByteChannel)},
+         * the stream will be buffered with this size of buffer (default: 100MiB).
          */
         public UnpackerConfig withBufferSize(int bytes)
         {
--- a/msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
@@ -518,7 +518,7 @@ public class MessagePack
 
         private int stringSizeLimit = Integer.MAX_VALUE;
 
-        private int bufferSize = 100 * 1024 * 1024;
+        private int bufferSize = 8192;
 
         private int stringDecoderBufferSize = 8192;
 
@@ -745,7 +745,7 @@ public class MessagePack
 
         /**
          * When a packer is created with {@link #newUnpacker(InputStream)} or {@link #newUnpacker(ReadableByteChannel)},
-         * the stream will be buffered with this size of buffer (default: 100MiB).
+         * the stream will be buffered with this size of buffer (default: 8192).
          */
         public UnpackerConfig withBufferSize(int bytes)
         {
--- a/msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
@@ -1326,38 +1326,27 @@ public class MessageUnpacker
      *
      * @return the size of the array to be read
      * @throws MessageTypeException when value is not MessagePack Array type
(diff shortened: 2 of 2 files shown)


msgpack-core/src/main/java/org/msgpack/core/MessagePack.java
```diff
@@ -518,7 +518,7 @@ public boolean isStr8FormatSupport()
 
         private int stringSizeLimit = Integer.MAX_VALUE;
 
-        private int bufferSize = 8192;
+        private int bufferSize = 100 * 1024 * 1024;
```
xerial, line 521: Don't change the default buffer size. msgpack-java is frequently used even for the small size of data.

xerial: Especially, creating a large buffer is not free as Java will use CPU resources for filling zeros to the array contents.

xerial (changes requested):
- Do not change the default buffer size from 8K
- Setting the default buffer size from config is ok
- Do not throw an exception in unpackXXXheader. Only readPayload needs to be fixed to support large volumes of data
- readPayload should swap the existing buffer to a large buffer if the payload length exceeds the current bufferSize

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
```diff
@@ -1324,27 +1326,38 @@ public Instant unpackTimestamp(ExtensionTypeHeader ext) throws IOException
...
+                }
             }
         }
-        throw unexpected("Array", b);
+
+        if (len > bufferSize) {
+            throw new MessageSizeException(String.format("cannot unpack a Array of size larger than %,d: %,d", bufferSize, len), len);
+        }
+
+        return len;
```
xerial, line 1360: Throwing error here is unnecessary if readPayload checks the buffer size.

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
```diff
@@ -1357,27 +1370,38 @@ public int unpackArrayHeader()
...
+                }
             }
         }
-        throw unexpected("Map", b);
+
+        if (len > bufferSize / 2) {
+            throw new MessageSizeException(String.format("cannot unpack a Map of size larger than %,d: %,d", bufferSize / 2, len), len);
+        }
+
+        return len;
```
xerial, line 1404: Unnecessary change for the same reason

msgpack-core/src/main/java/org/msgpack/core/MessageUnpacker.java
```diff
@@ -1635,11 +1659,16 @@ public void readPayload(byte[] dst)
      *
      * @param length number of bytes to be read
      * @return the new byte array
+     * @throws MessageSizeException when length is larger than bufferSize
      * @throws IOException when underlying input throws IOException
      */
     public byte[] readPayload(int length)
             throws IOException
     {
+        if (length > bufferSize) {
```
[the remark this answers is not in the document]
napo0703: I'm sorry that I don't quite understand here.
Am I correct that the purpose of `UnpackerConfig.bufferSize` is to limit the size of the new byte array in `readPayload(int)`?
Do I need to change this code?

napo0703:
Buffer size limits now work, but some tests no longer pass.
Should I change the Unpacker buffer size used for this test in the settings?
```
-        val unpacker = MessagePack.newDefaultUnpacker(new InputStreamBufferInput(new ByteArrayInputStream(out.toByteArray)))
+        val unpacker = new MessagePack.UnpackerConfig().withBufferSize(16 * 1024).newUnpacker(new InputStreamBufferInput(new ByteArrayInputStream(out.toByteArray)))
```

xerial:
Upon re-reading the PR, I noticed some misunderstandings.

UnpackerConfig.bufferSize is used to control the internal buffer size, but it is not meant to limit the size of the input uncompressed data. In order to prevent loading excessively large data into memory, we need to introduce another configuration. 

I will close this pull request because it has become outdated. The requirement to limit the uncompressed data size and fail the Unpacker is still valid, but it should be addressed in a different context.

PR #667, opened 2022-07-12
adding a method writePayload(String src) 

nathan-sigsci:
This new method copies the lower 8-bits of the characters.  A separate writePayload(String src) method would be more performant because we're able to directly copy the bytes from the string to the right location in MessageBuffer.  Introducing encoding in an attempt to re-use the existing writePayload(byte[] src) creates unnecessary overhead which in turn will affect performance.

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
+++ b/msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
@@ -1182,4 +1182,25 @@ public class MessagePacker
         }
         return this;
     }
+
+    /**
+     * This method copies the lower 8-bits of the characters.
+     * A separate writePayload(String src) method would be more performant because we're able to
+     * directly copy the bytes from the string to the right location in MessageBuffer.  Introducing
+     * encoding in an attempt to re-use the existing writePayload(byte[] src) creates unnecessary overhead
+     * which in turn will affect performance.
+     *
+     * @param src the string data to add
+     * @return this
+     * @throws IOException
+     */
+
+    public MessagePacker writePayload(String src) throws IOException
+    {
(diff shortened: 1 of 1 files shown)


msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
```diff
@@ -1129,6 +1129,35 @@ public MessagePacker writePayload(byte[] src, int off, int len)
...
+     * which in turn will affect performance.
+     *
+     * @param src the string data to add
+     * @return this
+     * @throws IOException
+     */
+
+    public MessagePacker writePayload(String src) throws IOException
+    {
+        int len = src.length();
```
xerial, line 1146: If the input string contains non-ascii characters, src.getBytes() will produce a longer array than src.length(), so using this method can easily create corrupted data.

msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java
```diff
@@ -1129,6 +1129,35 @@ public MessagePacker writePayload(byte[] src, int off, int len)
...
+        if (buffer == null || buffer.size() - position < len || len > bufferFlushThreshold) {
+            flush();  // call flush before write
+            // Directly write payload to the output without using the buffer
+            out.write(src.getBytes(), 0, len);
```
xerial, line 1150: If calling src.getBytes() is allowed here, you should be able to use addPayload(byte[]) instead.

xerial:
Hi @nathan-sigsci,

Thank you for creating a PR, but the intended optimization is already available as .addPayload method. You can use these method instead. 

I'll close the PR.

Issue #691, opened 2022-11-16
Improve memory usage via reuse InputStreamBufferInput


koo-taejin:
I am testing to use MsgPack instead of Jackson in Spring.
It was confirmed that MsgPack consumes more memory than Jackson.

I dug into that issue, it was confirmed that a lot of memory is created when creating `InputStreamBufferInput`.
So I have make code very similar to the previous reuse form.

After the change, I had confirmed that the memory usage was greatly reduced, and the slope of the heap memory was also changed to be very stable.

I didn't add a test because I don't know whether you merge the code or not, but if you are going to accept it, I'll add a simple test code.

- as-is 
![msgpack as-is](https://user-images.githubusercontent.com/10057874/202207778-9a1b3cfd-d206-4957-8e2b-f3509f1f5f43.JPG)

- to-be 
![msgpack to be ](https://user-images.githubusercontent.com/10057874/202207802-80e6d746-5563-4a14-8aca-0a1287464605.JPG)

Thanks :)

komamitsu:
@koo-taejin Great suggestion. Thanks for looking into the issue. And sorry for the late reply.
I looked at https://github.com/koo-taejin/msgpack-java/commit/7e59e496fc1c233d06f0172ceaceefd100469017 and the basic idea seems good. But I think there is room to improve it as follows:
1. This MessageBufferInput (InputStreamBufferInput) cache can be treated as a part of existing `reuseResourceInParser` feature. So, we can avoid introducing the new feature flag `reuseInputStreamBufferInput`.
2. Maintaining the different cache lifetimes of MessageUnpacker and MessageBufferInput would increase the complexity. We should keep using the same ThreadLocal to hold the new resource MessageBufferInput by change the value type from `Tuple<>` to `Triple<Object, MessageUnpacker, MessageBufferInput>`. I think this requires to modify some existing code.

If you want to move forward on this improvement and create a PR, I hope the above idea sounds reasonable to you. If you're busy or not interested in creating a PR, I think I can take over this performance improvement as I have some time now.

koo-taejin:
@komamitsu 

Please check this PR is what you want. 
If it is right, then I am going to add test code. 
And If you want to change my code, then you can change it viamaintainer option. 

Thanks :)

komamitsu:
[...]

PR #692, opened 2022-11-16
[#691] Improve memory usage via reuse InputStreamBufferInput

koo-taejin:


xerial:
Thank you for the PR. 

msgpack-java is commonly used for encoding and decoding relatively small amounts of data. The JVM has been optimized for frequent memory allocations, so memory optimization like this is usually unnecessary. 

I will close the PR as the CI is also not passing.

xerial:
#704 would be a following work

PR #745, opened 2023-07-10
Add `MessagePackMapper#handleBigDecimalAsString`

komamitsu:
## Context

We've received some issues that `jackson-dataformat-msgpack` fails to serialize large BigDecimal values. It's because the default internal type `double` can't store large BigDecimal values. We've found [a workaround](https://github.com/msgpack/msgpack-java/blob/develop/msgpack-jackson/README.md#serialize-and-deserialize-bigdecimal-as-str-type-internally-in-messagepack-format), but the same issue has been reported.

## Changes

This PR introduces `MessagePackMapper#handleBigDecimalAsString` with which users can easily use the workaround, and recommends to use the new method in README


xerial:
@komamitsu It seems a similar type escalation mechanism is necessary for BigInteger as well

komamitsu:
Handled BigInteger in https://github.com/msgpack/msgpack-java/pull/768

PR #770, opened 2023-10-15
use `setup-java` instead of deprecated `olafurpg/setup-scala`

xuwei-k:
https://github.com/olafurpg/setup-scala/releases/tag/v14

image


xerial:
Thanks! And good to know that setup-scala has been deprecated

PR #785, opened 2023-12-30
Fix fixint test in MessagePackTest.

fangzheng:
According to [msgpack format spec](https://github.com/msgpack/msgpack/blob/master/spec.md#formats), positive fix int ranges from 0x00 to 0x7f. The test only covers from 0x00 to 0x79.


xerial:
Thanks. Merged

PR #786, opened 2023-12-30
Fix bug in ImmutableTimestampValueImpl.

fangzheng:
`ImmutableTimestampValueImpl.getData()` shifts `nsec` (which is a 32-bit `int`) by 34 bits.


xerial:
Thanks. Merged

PR #787, opened 2023-12-30
Remove dead code in MessagePacker.

fangzheng:
As the comment says, the check does nothing and can be removed.


xerial:
Thanks. Merged

PR #794, opened 2024-01-19
Fix ClassCastException when array values is set as List

takezoe:
```
java.lang.ClassCastException: class [Ljava.lang.Object; cannot be cast to class [Lorg.msgpack.value.Value; ([Ljava.lang.Object; is in module java.base of loader 'bootstrap'; [Lorg.msgpack.value.Value; is in unnamed module of loader 'app')
        at org.msgpack.value.Variable$ArrayValueAccessor.array(Variable.java:890)
        at org.msgpack.value.Variable$ArrayValueAccessor.immutableValue(Variable.java:851)
        at org.msgpack.value.Variable$ArrayValueAccessor.immutableValue(Variable.java:832)
        at org.msgpack.value.Variable.immutableValue(Variable.java:1136)
        at org.msgpack.value.Variable.toJson(Variable.java:1161)
```

### DIFF
--- a/msgpack-core/src/main/java/org/msgpack/value/Variable.java
+++ b/msgpack-core/src/main/java/org/msgpack/value/Variable.java
@@ -817,7 +817,7 @@ public class Variable
     {
         this.type = Type.LIST;
         this.accessor = arrayAccessor;
-        this.objectValue = v.toArray();
+        this.objectValue = v.toArray(new Value[v.size()]);
         return this;
     }
 


msgpack-core/src/main/java/org/msgpack/value/Variable.java
```diff
@@ -817,7 +817,7 @@ public Variable setArrayValue(List<Value> v)
     {
         this.type = Type.LIST;
         this.accessor = arrayAccessor;
-        this.objectValue = v.toArray();
```
takezoe, line 820: `ClassCastException` occurs here:
https://github.com/msgpack/msgpack-java/blob/4bdaf8e35f130e561ca964b578bbe696c4d1dd9f/msgpack-core/src/main/java/org/msgpack/value/Variable.java#L888-L891

msgpack-core/src/main/java/org/msgpack/value/Variable.java
```diff
@@ -817,7 +817,7 @@ public Variable setArrayValue(List<Value> v)
...
+        this.objectValue = v.toArray(new Value[v.size()]);
```
takezoe, line 820: `ClassCastException` occurs here:
https://github.com/msgpack/msgpack-java/blob/4bdaf8e35f130e561ca964b578bbe696c4d1dd9f/msgpack-core/src/main/java/org/msgpack/value/Variable.java#L888-L891

PR #800, opened 2024-02-14
Make the MessagePackParser#type field public

### LINKED ISSUE #799: Make the MessagePackParser#type field public
https://github.com/msgpack/msgpack-java/blob/fd386138fec6f2a469516dcb2b113dc66e6247e8/msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackParser.java#L67-L71

It's needed us to determine whether we have extension value or byte array in deserialazer.

ArtDu:
It's useful when you need to distinguishe bytes array from extension type

Closes #799

### DIFF
--- a/msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackParser.java
+++ b/msgpack-jackson/src/main/java/org/msgpack/jackson/dataformat/MessagePackParser.java
@@ -64,11 +64,11 @@ public class MessagePackParser
     private final IOContext ioContext;
     private ExtensionTypeCustomDeserializers extTypeCustomDesers;
 
-    private enum Type
+    public enum Type
     {
         INT, LONG, DOUBLE, STRING, BYTES, BIG_INT, EXT
     }
-    private Type type;
+    public Type type;
     private int intValue;
     private long longValue;
     private double doubleValue;


komamitsu:
@ArtDu Can you elaborate your use case?

ArtDu:
@komamitsu 

> @ArtDu Can you elaborate your use case?

We use extensions values and byte arrays as well to store data in Tarantool(https://en.wikipedia.org/wiki/Tarantool).
And we can distinguishe those two types in jackson embedded object only by class type and it's no elegant way

```java
        private boolean isExtensionValue(Object object) {
            return object instanceof MessagePackExtensionType;
        }


        @Override
        public Object deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
            if (p.currentTokenId() != JsonTokenId.ID_EMBEDDED_OBJECT || !isExtensionValue(p.getEmbeddedObject())) {
                return super.deserialize(p, ctxt);
            }
            MessagePackExtensionType ext = p.readValueAs(MessagePackExtensionType.class);
            byte type = ext.getType();
            ...
        }
```

Instead of it we can use type variable that is stored in MessagePackParser
```java
 @Override
        public Object deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
            MessagePackParser mpParser = (MessagePackParser)p;
            if (mpParser.getType().equals(Type.EXT)) {
                return super.deserialize(p, ctxt);
            }
            ...
        }

```

komamitsu:
@ArtDu Thanks for sharing the details.

`jackson-dataformat-msgpack` has a function to deserialize extention type values by registering the target ext-type ID  and callback lambda using [org.msgpack.jackson.dataformat.ExtensionTypeCustomDeserializers#addCustomDeser](https://github.com/msgpack/msgpack-java/tree/main/msgpack-jackson#deserialize-extension-type-value-directly). Does it work for your use case?

BTW, as you use jackson-databind interface, I guess you serialize Java objects to MessagePack format data and/or deserialize MessagePack format data to Java object vise verse. So, I think you know all the type of values when deserializing MessagePack format data since you know how the data is serialized. For instance, let's say you serialize an instance of this class

```java
public static class Foo {
  String name;
  byte[] data;
}
```

and then, you know the type of `data` is stored as a byte array and don't need to consider extension types in this case. What do you think?

ArtDu:
@komamitsu 

> `jackson-dataformat-msgpack` has a function to deserialize extention type values by registering the target ext-type ID and callback lambda using [org.msgpack.jackson.dataformat.ExtensionTypeCustomDeserializers#addCustomDeser](https://github.com/msgpack/msgpack-java/tree/main/msgpack-jackson#deserialize-extension-type-value-directly). Does it work for your use case?

In some cases it works for me, but I also need a serializer. That's why I use `ObjectMapper(...).registerModule()` method.

Also `addCustomDeser` way can't set targetType. 
Sometimes I need to use a mapper in this way `mapper.readValue(bytes, SpecialClass.class)`.
For example, we can interpret ExtensionValue (Datetime) as ZonedDateTime, as Instant, or as LocalTime. In this way, we determine how to get the needed type.


> BTW, as you use jackson-databind interface, I guess you serialize Java objects to MessagePack format data and/or deserialize MessagePack format data to Java object vise verse. So, I think you know all the type of values when deserializing MessagePack format data since you know how the data is serialized. For instance, let's say you serialize an instance of this class
> 
> ```java
> public static class Foo {
>   String name;
>   byte[] data;
> }
> ```
> 
> and then, you know the type of `data` is stored as a byte array and don't need to consider extension types in this case. What do you think?

We have two types of interaction:
1. When we expect an explicit class that we'll read from the database (msgpack bytes). It's like I described above
2. When we don't know what type we'll come and we need to read bytes by its msgpack own type. The type could be an extension or raw bytes (And at this point I want to determine by the `Type`)

komamitsu:
@ArtDu I think understand your use case. Honestly speaking, the usage seems a bit too dynamic in terms of type-safety and to not mach jackson-databind interface. I don't think it's a common use case, and I'm hesitating to approve the change.

As for `org.msgpack.jackson.dataformat.MessagePackParser#type`, it's not supposed to be externally exposed and I might change it's behavior in the future. If it's exposed by this PR, it would be a possible maintenance concern.

Regarding the sample code you shared, I think you know the value is an embedded object at least in `deserialize()` and you can simplify it w/o checking `currentTokenId` like this.

```java
        @Override
        public Object deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
            // At this point, the current value must be an embedded object, right?
            Object embedded = p.getEmbeddedObject();
            if (!(embedded instanceof MessagePackExtensionType)) {
                // This should be `byte[]`.
                handleByteArrayAsXxxxx(embedded);
                return;
            }
            MessagePackExtensionType ext = (MessagePackExtensionType) embedded;
            byte type = ext.getType();
            ...
        }
```

ArtDu:
This w/a works fine from experience. Since that I close this PR and ticket 

> We use extensions values and byte arrays as well to store data in Tarantool(https://en.wikipedia.org/wiki/Tarantool).
> And we can distinguishe those two types in jackson embedded object only by class type and it's no elegant way
> 
> ```java
>         private boolean isExtensionValue(Object object) {
>             return object instanceof MessagePackExtensionType;
>         }
> 
> 
>         @Override
>         public Object deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
>             if (p.currentTokenId() != JsonTokenId.ID_EMBEDDED_OBJECT || !isExtensionValue(p.getEmbeddedObject())) {
>                 return super.deserialize(p, ctxt);
>             }
>             MessagePackExtensionType ext = p.readValueAs(MessagePackExtensionType.class);
>             byte type = ext.getType();
>             ...
>         }
> ```

PR #986, opened 2026-05-27
Add module name to the package manifest attributes

thatcomputerguy0101:
I tried pulling `msgpack-core` in through Maven, and I noticed that the module name was being parsed from the jar file name as `msgpack.core` instead of `org.msgpack.core`. This change adds the module name defined in the sbt config to the package manifest attributes so that Maven autodetects the correct name.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -21,10 +21,6 @@ ThisBuild / scmInfo  :=
     )
   )
 
-ThisBuild / packageOptions += Package.ManifestAttributes(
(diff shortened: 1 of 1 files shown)


thatcomputerguy0101:
(PS: I'm not very familiar with sbt, so there may be a better way to do this.)

thatcomputerguy0101:
~~Actually, I failed to test my final revision of this properly, so this doesn't work right now.~~ I think it is fixed now.

PR #987, opened 2026-05-31
Add msgpack-jackson3 module for Jackson 3.x support

### LINKED ISSUE #933: Add Jackson 3 support
I'd be great to have a new module that is compatible with Jackson 3 API.

komamitsu:
## Summary

- Adds a new `msgpack-jackson3` submodule (`jackson-dataformat-msgpack3`) that ports the existing Jackson 2.x integration to Jackson 3.x (`tools.jackson` namespace)
- Requires Java 17+ and Jackson 3.1.2; the module is excluded from the build on older JDKs so existing Java 8 users are unaffected
- Includes JMH benchmarks (`src/jmh/java`) runnable via `./sbt "msgpack-jackson3/jmh:run"`
- Fixes https://github.com/msgpack/msgpack-java/issues/933

## Key differences from `msgpack-jackson`

- Package namespace: `com.fasterxml.jackson` → `tools.jackson`
- `TokenStreamLocation` replaces `JsonLocation`; byte offset passed as `columnNr`
- Lightweight `MessagePackWriteContext` replaces `SimpleStreamWriteContext`

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -28,6 +28,7 @@ jobs:
               - 'project/build.properties'
               - 'msgpack-core/**'
               - 'msgpack-jackson/**'
+              - 'msgpack-jackson3/**'
             docs:
               - '**.md'
               - '**.txt'
@@ -65,6 +66,16 @@ jobs:
           key: ${{ runner.os }}-jdk${{ matrix.java }}-${{ hashFiles('**/*.sbt') }}
           restore-keys: ${{ runner.os }}-jdk${{ matrix.java }}-
       - name: Test
-        run: ./sbt test
+        run: |
+          if [[ ${{ matrix.java }} -lt 17 ]]; then
+            ./sbt msgpack-core/test msgpack-jackson/test
+          else
+            ./sbt test
+          fi
       - name: Universal Buffer Test
-        run: ./sbt test -J-Dmsgpack.universal-buffer=true
+        run: |
+          if [[ ${{ matrix.java }} -lt 17 ]]; then
+            ./sbt msgpack-core/test msgpack-jackson/test -J-Dmsgpack.universal-buffer=true
+          else
+            ./sbt test -J-Dmsgpack.universal-buffer=true
+          fi
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,7 +30,17 @@ jobs:
         env:
           PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
         run: |
-          ./sbt publishSigned
+          ./sbt msgpack-core/publishSigned msgpack-jackson/publishSigned
+      # msgpack-jackson3 requires JDK 17+
+      - uses: actions/setup-java@v5
+        with:
+          java-version: 17
+          distribution: temurin
+      - name: Build bundle for msgpack-jackson3
+        env:
+          PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
+        run: |
+          ./sbt msgpack-jackson3/publishSigned
       - name: Release to Sonatype
         env:
           SONATYPE_USERNAME: '${{ secrets.SONATYPE_USERNAME }}'
--- a/README.md
+++ b/README.md
@@ -53,8 +53,16 @@ For using DirectByteBuffer (off-heap memory access methods) in JDK17, you need t
 
 ### Integration with Jackson ObjectMapper (jackson-databind)
 
-msgpack-java supports serialization and deserialization of Java objects through [jackson-databind](https://github.com/FasterXML/jackson-databind).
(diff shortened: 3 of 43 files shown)


komamitsu:
@xerial Can you take a quick look at the changes of `.github/workflows/*` and `*.sbt`?

xerial:
@komamitsu Do you still need to maintain msgpack-jackson for Jackson 2.x? If we don't have critical use cases, I'll lean to dropping Jackson 2 support as supporting both Jackson2 and 3 will be overkill for msgpack-java project.

xerial:
current usage: https://mvnrepository.com/artifact/org.msgpack/jackson-dataformat-msgpack/used-by

komamitsu:
@xerial Jackson 2 is still actively used, and Jackson 3 requires Java 17+, so not everyone can migrate yet. I think we should support both for a while.

hiSandog:
The split module is a pragmatic way to support Jackson 3 without raising the whole project baseline. One release detail to double-check: the CI/release workflow now publishes Jackson 2 artifacts on the original JDK and `msgpack-jackson3` after switching to JDK 17, but consumers will need a clear artifact-name mapping in the top-level README/release notes. Calling out `jackson-dataformat-msgpack` for Jackson 2 vs `jackson-dataformat-msgpack-jackson3` for Jackson 3 would reduce upgrade confusion.

komamitsu:
@hiSandog Good point. Addressed in https://github.com/msgpack/msgpack-java/pull/987/commits/531d3d0e31a1ae8a9ffd78e762e08fb920e35512

komamitsu:
@xerial I addressed the conflict, PTAL.

BTW, it seems `Universal Buffer Test` has been [skipped](https://github.com/msgpack/msgpack-java/actions/runs/28471835766/job/84386267691) since https://github.com/msgpack/msgpack-java/pull/998 was merged.

xerial:
@komamitsu Thank you for this PR — the Jackson 3 port itself looks solid, and I agree with your point that we should support both Jackson 2 and 3 for a while, since Jackson 3 requires Java 17+ and many users can't migrate yet.

Before merging, I'd like to propose restructuring the module naming. My concern with `jackson-dataformat-msgpack3` is long-term: if Jackson 3 takes over Jackson 2 in the next couple of years, the unmarked name `jackson-dataformat-msgpack` will be stuck on the legacy version, and Jackson 3 users (the growing cohort) would eventually need a second migration from `-msgpack3` back to the plain name. I'd rather put the clean name on the future from day one.

Concretely, I propose we cut **msgpack-java 1.0.0** with this structure:

- **`jackson-dataformat-msgpack` (1.x) = Jackson 3**: your Jackson 3 code from this PR becomes the contents of the `msgpack-jackson` module. To avoid a classpath clash with the Jackson 2 artifact, the Java package moves to **`org.msgpack.jackson3.dataformat`**. Since Jackson 3 adopters are already rewriting every `com.fasterxml.jackson` import to `tools.jackson`, one more changed import costs them almost nothing — and it means both artifacts can coexist in one application for incremental migration (which was a design goal of Jackson's own `tools.jackson` rename).
- **`jackson2-dataformat-msgpack` (1.x) = Jackson 2**: the current `msgpack-jackson` code moves here unchanged, keeping the `org.msgpack.jackson.dataformat` package. Jackson 2 users who want to keep receiving msgpack-core updates only need a one-line artifactId change — no import changes at all. This module would be maintenance-mode (Jackson 2.x bumps and bug fixes only), documented as such in the README, to be dropped in a future major release when Jackson 2 usage fades.
- **0.9.x freezes**: users who do nothing stay on `jackson-dataformat-msgpack` 0.9.10 and nothing breaks. We'll keep a `0.9.x` branch as an escape hatch for critical fixes, but the plan is zero releases from it — this avoids maintaining two release lines with msgpack-core backports.

If dependabot/renovate proposes 0.9.x → 1.0.0 to a Jackson 2 user, it fails loudly at compile time (the old package is gone from that artifact), and the release notes will point to `jackson2-dataformat-msgpack` — so there's no silent breakage.

Since msgpack-java's API has been stable for a decade while sitting at 0.x, I think this transition is also the right occasion to finally signal 1.0.0.

Mechanically this reuses essentially all of your work — the Jackson 3 sources just land in `msgpack-jackson` with the package rename, and the JDK-conditional build exclusion you added carries over (the module still requires Java 17+ while msgpack-core and the Jackson 2 module stay on Java 8). The OSGi bundle symbolic names and export packages would be updated to match.

WDYT? If you're on board, I'm happy to help with the build.sbt / workflow changes for the restructuring.

xerial:
I've opened #1012 implementing the restructuring proposed above, built on top of this PR's commits (so all the Jackson 3 port work is preserved and attributed). @komamitsu PTAL — if you prefer this direction we can continue the review there; if not, happy to discuss further here.

While enabling the module restructuring I also found that the Jackson 2 module's JUnit 5 tests have been silently skipped on main (no Jupiter test interface in the sbt build — likely related to the skipped Universal Buffer Test you noticed), and fixing that surfaced a small `JsonLocation` regression from #903. Both are fixed in #1012.

